PR #51, opened 2026-05-01
Strengthen deterministic safety rings: runtimeClass, egress allowlist, Cedar policy, Rule of Two, code scanner

### LINKED ISSUE #42: Strengthen deterministic safety rings: runtimeClass, egress allowlist, Cedar policy engine, Rule of Two, code scanner
## Summary

Strengthen the *deterministic* safety rings of stirrup so they more closely match the layered defence-in-depth blueprint in `stirrup-dod.md`. The blueprint is explicit that LLM-based guards (Llama Prompt Guard 2, Granite Guardian) are probability reducers, not gates: blast-radius bounds come from rings 1 (hard isolation), 2 (capability-scoped tools + policy engine), and 5 (deterministic output controls). This issue covers all three. The probabilistic ML guards (rings 3 + 4) are tracked separately.

## Motivation

The 2026 SoK on prompt-injection defences (arXiv 2601.17548) and Nasr et al.'s *"The Attacker Moves Second"* (October 2025) both show adaptive attacks defeat any single classifier with >85–90% ASR. Meta's *"Agents Rule of Two"* (October 2025) and DeepMind's CaMeL paper (arXiv:2503.18813) jointly argue the only reliable controls are deterministic: sandboxing, capability tokens, policy engines, and structural information-flow constraints. stirrup already has partial coverage; this issue closes the gaps that don't depend on running an external model.

## What stirrup already has

- Container executor (`harness/internal/executor/container.go`) with `CapDrop:[ALL]`, `no-new-privileges`, `NetworkMode:none` by default. Resource limits are wired through.
- Permission policies (`harness/internal/permission/{allowall,denysideeffects,askupstream}.go`) with `WorkspaceMutating` and `RequiresApproval` flags on each tool (closed in #33).
- Read-only mode invariants in `types/runconfig.go::ValidateRunConfig` — read-only modes must use a restrictive policy and cannot enable mutating tools.
- Dynamic-context sanitisation (`harness/internal/security/dynamiccontext.go`) wrapping untrusted input in `<untrusted_context>` tags and stripping nested delimiters.

## Gaps closed by this issue

1. The container executor has no kernel-isolation tier — Linux user-namespace caps only.
2. `NetworkConfig.Mode == "allowlist"` in `types/runconfig.go:185-189` is **silently weakened** to plain `bridge` networking at `harness/internal/executor/container.go:67-73` ("Phase 1: allowlist uses bridge networking"). Allowlists are not enforced anywhere.
3. There is no policy engine for tool-call authorisation. `deny-side-effects` and `ask-upstream` are coarse-grained per-tool toggles; there is no way to say "allow `run_command` only when `cmd` does not contain `rm -rf` or env-secrets" or "allow `web_fetch` only to allowlisted hosts" — the rules the blueprint lists in Ring 2.
4. The Meta "Rule of Two" — never let a session simultaneously hold (a) untrusted input, (b) sensitive data, (c) external side-effects — is not encoded anywhere. A `research`-mode run with a user-provided URL plus secrets in env plus `web_fetch` enabled is currently legal.
5. There is no static analysis of code the agent writes. `write_file` and `edit_file` succeed regardless of whether the content contains hardcoded secrets, eval/exec sinks, or known-insecure patterns.

## Scope

In scope:

- **B1** runtimeClass option for the container executor (gVisor / Kata Containers).
- **B2** real FQDN egress allowlist enforcement, replacing the "bridge" fallback.
- **B3** new `policy-engine` PermissionPolicy implementation backed by Cedar.
- **B4** Rule-of-Two invariant in `ValidateRunConfig`.
- **B5** new `CodeScanner` hook on the edit pipeline (Semgrep + Code Shield).

Out of scope (tracked elsewhere):

- LLM-based prompt-injection or content guards (Rings 3 + 4) — separate issue.
- CaMeL dual-LLM information-flow control — explicitly deferred; conflicts with `spawn_agent` recursion.
- External Invariant-style gateway sidecar — explicitly deferred.
- Egress DPI on the proxy itself — blueprint scopes this to a later round (`stirrup-dod.md` ring-1 caveat).

### B1 — runtimeClass for the container executor

Add an optional `Runtime` field to `ContainerExecutorConfig` (`harness/internal/executor/container.go:23-29`) and surface it on `types.ExecutorConfig` as a new `Runtime string` (values: `""` (default), `"runc"`, `"runsc"` (gVisor), `"kata"`, `"kata-qemu"`, `"kata-fc"`).

The Docker Engine API accepts a `Runtime` field on `HostConfig` (`POST /containers/create`). Plumb it through `executor/container_api.go::hostConfig`. Validate in `ValidateRunConfig` against a closed set; reject unknown runtimes. Surface as `--container-runtime` flag and in `examples/runconfig/full.json`. Document required host setup in `docs/sandbox.md` (gVisor needs `runsc` registered with the Docker daemon; Kata needs `kata-runtime`).

This is purely opt-in: existing deployments default to runc with no behaviour change. K8s users set `runtimeClass` on the pod spec instead, so this matters only for the local-Docker container executor.

### B2 — FQDN egress allowlist

Today `Network.Mode == "allowlist"` falls back to `bridge` (no enforcement). Replace this with a real allowlist mechanism.

Approach: introduce a small in-process **egress proxy** per agent session, modelled on Anthropic's Claude Code sandboxing pattern (referenced in `stirrup-dod.md` ring-1 description). The proxy is an HTTP/SOCKS5 forward proxy that resolves the destination FQDN, matches it against `Network.Allowlist`, and only forwards if it matches. Implementation:

1. New package `harness/internal/executor/egressproxy/` exposing `func Start(allowlist []string) (*Proxy, error)` returning the listen address, a `Stop()` closer, and an emitter for `egress_blocked` security events.
2. Allowlist entries are FQDNs with optional leading `*.` for wildcard subdomains. No port mismatch is allowed: only 443 unless explicitly suffixed (`example.com:80`). DNS resolution happens inside the proxy; the container has no DNS access.
3. ContainerExecutor starts the proxy on the host network namespace, then sets the container's `NetworkMode` to a fresh user-defined bridge with a restrictive `iptables` chain that DROPs all traffic except to the proxy's listen address. The container env gets `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` set so well-behaved clients (curl, git over https, the language stdlib HTTP client) honour it.
4. `none` mode unchanged — no allowlist, no proxy, no networking.
5. Local executor (no container) refuses `Mode: allowlist` and returns a config error at construction time. The blueprint is clear that egress controls require a sandbox boundary.

Emit a `security_event` of type `egress_allowed` / `egress_blocked` per request through the OTel trace. Add unit tests against a fake upstream HTTP server and a real-DNS table-driven test for `*.example.com` matching.

Acknowledged limitation: HTTPS clients that bypass `HTTPS_PROXY` (raw TCP, custom DNS) will hit the iptables drop and surface as a connection error. That is the desired fail-closed behaviour.

### B3 — Cedar-backed `policy-engine` PermissionPolicy

Add a fourth `PermissionPolicy` type alongside `allow-all`, `deny-side-effects`, `ask-upstream`:

```go
// harness/internal/permission/policyengine.go
type PolicyEnginePolicy struct {
    cedar     *cedar.PolicySet
    fallback  PermissionPolicy // for tools the policy set does not cover
    security  SecurityEventEmitter
}
```

Use **Cedar** (`github.com/cedar-policy/cedar-go`, Apache 2.0, native Go, no Wasm or sidecar required) as the engine. Cedar's `principal/action/resource/context` model maps cleanly onto our `(subject, tool, input, context)` quadruple. OPA was the alternative but its native Go embedding (OPA SDK) is heavier and pulls Wasm/Rego runtimes; Cedar is a better fit for a single-binary harness.

Schema:

```cedar
// principal: User::"<runId>" (carries Mode, runId, capabilities)
// action:    Action::"tool:<name>"  e.g. Action::"tool:run_command"
// resource:  Tool::"<name>"
// context:   { input: ..., workspace: ..., dynamicContext: { ... } }

permit (
    principal in User::"any",
    action == Action::"tool:web_fetch",
    resource
) when {
    context.input.url like "https://*.github.com/*" ||
    context.input.url like "https://docs.python.org/*"
};

forbid (
    principal,
    action == Action::"tool:run_command",
    resource
) when {
    context.input.cmd like "*rm -rf*" ||
    context.input.cmd like "*chmod -R*" ||
    context.input.cmd like "*git push --force*"
};
```

Config schema (`types/runconfig.go::PermissionPolicyConfig`):

```go
type PermissionPolicyConfig struct {
    Type       string `json:"type"`        // adds "policy-engine"
    Timeout    int    `json:"timeout,omitempty"`
    PolicyFile string `json:"policyFile,omitempty"` // path to .cedar file
    Fallback   string `json:"fallback,omitempty"`   // type to consult on no-match (default: deny-side-effects)
}
```

Behaviour:

- `Check()` constructs the Cedar entity graph from the tool call, calls `cedar.IsAuthorized()`, returns `Allowed` for `Allow`, denies on `Forbid`, and consults the configured fallback policy on the no-match case (Cedar's "no decision").
- All decisions are logged via the existing `SecurityEventEmitter` with `policy_decision` / `policy_denied` events including the matched policy ID(s).
- Policy file is loaded once at construction and validated; bad files fail fast with a clear error.
- A small set of starter policies ships under `examples/policies/` covering: blocked destructive shell commands, GitHub-only `web_fetch`, no-secret-in-input check, `spawn_agent` capability-cap (no nested `run_command` in sub-agents).

### B4 — Rule of Two invariant

Add a structural validation in `types/runconfig.go::ValidateRunConfig`. Define three boolean session capabilities:

- `holdsUntrustedInput` — true if `len(DynamicContext) > 0` OR the prompt explicitly references a fetched URL OR `web_fetch` / `read_file` are enabled in a mode where the workspace contains externally-sourced content (heuristic: any mode other than `execution` with no git remote, plus any mode where a `--prompt-file` is provided from outside the workspace tree). v1 starts simple: `holdsUntrustedInput = len(DynamicContext) > 0 || webFetchEnabled || mcpServersEnabled`.
- `holdsSensitiveData` — true if any environment variable matching `*KEY*`, `*TOKEN*`, `*SECRET*`, `*PASSWORD*` is allowlisted into the executor environment OR the workspace contains a `.env`, `.npmrc`, `.netrc`, `~/.aws/credentials` etc. v1 starts simple: `holdsSensitiveData = anyAllowlistedSecretEnvVar || anySSMSecretRef`.
- `canCommunicateExternally` — true if `run_command` is enabled OR `web_fetch` is enabled OR any MCP server is configured OR the executor has any non-`none` `NetworkConfig`.

Rule: at most two of these may be true unless `PermissionPolicy.Type == "ask-upstream"`. If all three are true and the policy is not `ask-upstream`, reject the config with:

```
all three of {untrusted-input, sensitive-data, external-communication} cannot
hold simultaneously without the ask-upstream permission policy (Rule of Two)
```

Surface a structured warning (not error) in the trace at run start when *exactly* two hold, including which two. Add an opt-out `RuleOfTwo: { enforce: false }` for explicit override on operators who know what they're doing — but log loudly via a `security_event` of type `rule_of_two_disabled` so the override is auditable.

### B5 — `CodeScanner` hook on the edit pipeline

Add a small interface invoked after every successful `EditStrategy.Apply`:

```go
// harness/internal/security/codescanner.go
type CodeScanner interface {
    Scan(ctx context.Context, path string, content []byte) (*ScanResult, error)
}

type ScanResult struct {
    Findings []Finding
}

type Finding struct {
    Severity string // "block" | "warn"
    Rule     string
    Line     int
    Message  string
}
```

Implementations to ship:

1. **`semgrep`** — shells out to `semgrep --config auto --json` if `semgrep` is on `$PATH`. Skips silently if the binary is absent (with a single warning at startup).
2. **`patterns`** — pure-Go regex pack covering the top-10 hardcoded-secret patterns already in `harness/internal/security/logscrubber.go` (reuse those regexes), eval/exec sinks (`os.system`, `subprocess.*shell=True`, `eval(`), and obviously-malicious patterns. Always available.
3. **`composite`** — runs all configured scanners and unions their findings.

Wire it into the edit dispatch path in `harness/internal/edit/`. On a `block` finding, the edit returns an error that surfaces as a tool failure; on `warn`, it logs and emits a `code_scan_warning` security event but the edit succeeds. Configurable in RunConfig:

```go
type CodeScannerConfig struct {
    Type        string   `json:"type"`            // "none" | "patterns" | "semgrep" | "composite"
    Scanners    []string `json:"scanners,omitempty"` // for composite
    BlockOnWarn bool     `json:"blockOnWarn,omitempty"`
}
```

Default for `execution` mode: `patterns`. Default for read-only modes: `none` (no edits happen anyway).

## File-level impact

- `types/runconfig.go` — add `Runtime` to `ExecutorConfig`; extend `PermissionPolicyConfig`; new `RuleOfTwoConfig` + `CodeScannerConfig`; new validations.
- `proto/harness/v1/harness.proto` — mirror the new fields. Run `buf generate` after edits.
- `harness/internal/executor/container.go`, `container_api.go` — runtimeClass plumbing; egress proxy lifecycle; Network.Mode handling.
- `harness/internal/executor/egressproxy/` — new package.
- `harness/internal/permission/policyengine.go` (+ test) — new policy.
- `harness/internal/permission/permission.go` — extend constructor switch.
- `harness/internal/security/codescanner.go`, `codescanner_patterns.go`, `codescanner_semgrep.go`, `codescanner_composite.go` (+ tests).
- `harness/internal/edit/multi.go` — call out to scanner after successful apply.
- `harness/internal/core/factory.go` — wire `CodeScanner` and the new policy.
- `harness/cmd/stirrup/cmd/harness.go` — add `--container-runtime`, `--permission-policy-file`, `--code-scanner` flags.
- `examples/runconfig/full.json` — populate every new field.
- `examples/policies/` — new directory with starter Cedar policies.
- `docs/sandbox.md` — new doc covering runtimeClass setup, egress proxy semantics, policy authoring, Rule-of-Two, scanner config.
- `CLAUDE.md` — short summary of the new components in the architecture section.

## Acceptance criteria

- `go test ./harness/... ./types/... ./eval/...` passes.
- New unit tests cover: Cedar `Allow` / `Forbid` / `NoDecision` paths; egress proxy allowlist matching incl. wildcards; runtimeClass passes through to the create-container request; Rule-of-Two validation rejects the all-three case and warns on the two-of-three case; pattern scanner blocks on a planted secret and a planted `eval(` sink.
- Integration test that boots a containerExecutor with `Network.Mode: allowlist`, allowlist `["api.github.com"]`, and verifies `curl https://api.github.com` works inside the container while `curl https://example.org` is blocked.
- CI eval suite still passes against the existing baselines (`eval/baselines/`).
- `examples/runconfig/full.json` validates and runs end-to-end against a real Anthropic key with the new fields populated.
- `docs/sandbox.md` walks through the four configurations: dev (allow-all + runc + none), defaults (deny-side-effects + runc + patterns scanner), hardened (policy-engine + runsc + allowlist proxy + composite scanner), and read-only (ask-upstream + runc + no scanner).

## Notes for the implementer

- Cedar entity construction needs to be schema-stable across releases — pin a Cedar schema version and fail loudly on mismatch. The starter policies should ship alongside the schema they were authored against.
- The egress proxy is the most subtle piece. Use `goproxy` (BSD-3, mature) only if a from-scratch implementation feels risky — but the proxy is small (a few hundred lines of `net/http.Server` with a `Director` that filters CONNECTs by SNI for HTTPS). Either way, verify TLS SNI matches the requested host to defeat a tampered HOST header.
- Cedar policies are loaded **once** at boot. Don't add a hot-reload path until there's a clear ops use case; it complicates the audit story.
- The Rule-of-Two heuristic is intentionally crude in v1. The blueprint explicitly says this is a *guideline operationalised as an invariant*; we will refine the booleans based on real eval-suite signal.
- Do not add a `--no-rule-of-two` CLI flag. Force the override to be in the RunConfig file so it is reviewable in pull requests.
- Do not weaken `ValidateRunConfig` to a warn-only mode. The blueprint argues hard rejection is the point.

## References

- `stirrup-dod.md` — full blueprint, in particular Rings 1, 2, 5 and the "Rule of Two" sub-section.
- arXiv:2503.18813 (DeepMind, CaMeL).
- Meta, *"Agents Rule of Two"* (October 2025).
- arXiv 2601.17548 — 2026 SoK on prompt-injection defences.
- Cedar policy language — https://www.cedarpolicy.com/.
- Anthropic, *Claude Code sandboxing* — https://www.anthropic.com/engineering/claude-code-sandboxing.

rxbynerd:
Closes #42.

## Summary

Implements the five deterministic safety-ring sub-features (B1–B5) tracked in #42, layered defence-in-depth controls that don't depend on running an external classifier model.

- **B1 — runtimeClass for the container executor.** New `executor.runtime` config (`runc | runsc | kata | kata-qemu | kata-fc`) and `--container-runtime` flag, plumbed through Docker `HostConfig.Runtime`. Closed-set validation; empty = engine default.
- **B2 — FQDN egress allowlist.** New `harness/internal/executor/egressproxy/` package: a host-side HTTP forward proxy with FQDN/wildcard matching, default port 443, and TLS SNI verification (defeats tampered HOST headers; absent SNI is denied, not allowed). Container reaches it via `host.docker.internal` + `HTTP_PROXY`. Replaces the previous silent `Network.Mode == "allowlist"` → `bridge` weakening. Per-decision `egress_allowed`/`egress_blocked` security events. Local executor refuses `allowlist` mode at construction. **The host iptables drop is honestly deferred to v1.5** — v1 fails closed for cooperating clients only; documented in code and `docs/sandbox.md`.
- **B3 — Cedar-backed `policy-engine` PermissionPolicy.** New `harness/internal/permission/policyengine.go` using `github.com/cedar-policy/cedar-go` v1.6.0. Entity model: `User::"<runId>"` (with `mode`, `parentRunId`, capabilities); `Action::"tool:<name>"`; `Tool::"<name>"`; `context: { input, workspace, dynamicContext }`. Allow / Forbid / no-decision-falls-through-to-fallback. Bounded recursion (depth 64) on Cedar value construction to defend against deeply-nested tool inputs. Sub-agents get a child-scoped `PolicyEnginePolicy` with `parentRunId` set, so `subagent-capability-cap.cedar` actually constrains them. `policy_decision`/`policy_denied` security events surface matched policy IDs. Starter policies under `examples/policies/`: destructive-shell, github-only-fetch, no-secret-in-input, subagent-capability-cap.
- **B4 — Rule of Two invariant in `ValidateRunConfig`.** Three booleans (`untrusted-input`, `sensitive-data`, `external-comm`) — all-three rejected unless policy is `ask-upstream` or `RuleOfTwo.Enforce: false` is set. Override emits a `rule_of_two_disabled` security event at run start (auditable in PR review per the issue's requirement); two-of-three emits `rule_of_two_warning`. No `--no-rule-of-two` CLI flag (override must live in the RunConfig). Proto field is `optional bool enforce` (proto3 optional) so an empty `RuleOfTwoConfig` sub-message no longer silently disables enforcement.
- **B5 — `CodeScanner` hook on the edit pipeline.** New `harness/internal/security/codescanner/` (patterns / semgrep / composite / none) wrapped via `edit.ScannedStrategy`. Patterns reuses the canonical `logscrubber.go` regex set plus eval/exec sinks. Semgrep config path is configurable (default `auto` for back-compat; documented as making outbound HTTP). Block findings fail the edit and roll back the file; warn findings emit `code_scan_warning` and proceed. Defaults: `patterns` for execution mode, `none` for read-only modes.

Plus: new `--container-runtime`, `--permission-policy-file`, `--code-scanner` CLI flags. `examples/runconfig/full.json` populates every new field. New `docs/sandbox.md` walks through B1–B5 and four canonical configurations (dev / defaults / hardened / read-only). `CLAUDE.md` summarises the new components.

44 commits, 5 review reports (code, security, spec-compliance, test-coverage, api-design) consolidated into a remediation brief; all 8 must-fix and 8 should-fix items addressed in a final remediation pass. The remediation pass also caught and fixed a previously-silent gRPC translation gap where new fields were dropped between proto and `types.RunConfig`.

## Test plan

- [x] `go build ./harness/... ./types/... ./eval/...` — green
- [x] `go test ./harness/... ./types/... ./eval/...` — green (~1245 pass cases across 29 packages, +44 net new for remediation alone)
- [x] `TestExampleFullJSONLoadsAndValidates` — green with the populated example
- [x] `buf generate` — clean (cedar-go is the only new dep)
- [ ] Reviewer: spot-check the four canonical configurations in `docs/sandbox.md` validate cleanly against `ValidateRunConfig`
- [ ] Reviewer: confirm starter Cedar policies under `examples/policies/` parse and the entity model matches their expectations
- [ ] Reviewer: confirm the v1 egress-proxy threat model ("fail closed for cooperating clients only") is acceptable until the iptables follow-up lands
- [ ] Optional manual: run an integration test with `Network.Mode: allowlist` against a real `Docker Engine 20.10+` to confirm `host.docker.internal:host-gateway` plumbing and the proxy lifecycle

## Known limitations / follow-ups

- **iptables fail-closed enforcement** is deferred to v1.5 (issue body acknowledges this; documented in `harness/internal/executor/container.go` and `docs/sandbox.md`). v1 relies on `HTTP_PROXY` cooperation. Will be filed as a follow-up.
- A handful of nits (response-body size limit on plain HTTP; AKIA secret pattern; sub-agent IPv6 literals; Cedar schema-version mismatch behaviour; etc.) collected in the synthesizer's deferred list. Will be filed as a single nits follow-up.

### DIFF
--- a/harness/internal/executor/container_test.go
+++ b/harness/internal/executor/container_test.go
@@ -782,6 +782,80 @@ func TestContainerExecutor_NetworkModes(t *testing.T) {
 	}
 }
 
+func TestContainerExecutor_Runtime(t *testing.T) {
+	// Capture the raw POST body so we can assert exactly what bytes go on
+	// the wire. The Engine API treats a missing Runtime field as "use the
+	// daemon default", so it matters that we (a) include "Runtime":"runsc"
+	// when set and (b) omit the field entirely when not set — otherwise
+	// older Docker versions can fail with `runtime "" not registered`.
+	var rawBody []byte
+
+	sock, cleanup := mockEngineServer(t, map[string]http.HandlerFunc{
+		"POST /containers/create": func(w http.ResponseWriter, r *http.Request) {
+			rawBody, _ = io.ReadAll(r.Body)
+			w.Header().Set("Content-Type", "application/json")
+			_ = json.NewEncoder(w).Encode(containerCreateResponse{ID: "test-id"})
+		},
+		"POST /containers/*/start": func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusNoContent)
+		},
+		"POST /containers/*/stop": func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusNoContent)
+		},
+		"DELETE /containers/*": func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusNoContent)
+		},
+	})
+	defer cleanup()
+
+	// Configured runtime must appear in the create-container body.
+	exec, err := NewContainerExecutor(ContainerExecutorConfig{
+		Image:      "ubuntu:22.04",
+		HostDir:    "/tmp/workspace",
+		SocketPath: sock,
+		Runtime:    "runsc",
+	})
+	if err != nil {
+		t.Fatalf("NewContainerExecutor: %v", err)
+	}
+	_ = exec.Close()
+
+	if !strings.Contains(string(rawBody), `"Runtime":"runsc"`) {
+		t.Errorf("create body missing Runtime=runsc; got: %s", string(rawBody))
+	}
+
+	// Hardening defaults must still be present alongside the runtime.
+	var parsed containerCreateRequest
+	if err := json.Unmarshal(rawBody, &parsed); err != nil {
+		t.Fatalf("unmarshal create body: %v", err)
+	}
+	if parsed.HostConfig.Runtime != "runsc" {
+		t.Errorf("HostConfig.Runtime: got %q, want %q", parsed.HostConfig.Runtime, "runsc")
+	}
+	if parsed.HostConfig.NetworkMode != "none" {
+		t.Errorf("HostConfig.NetworkMode: got %q, want %q", parsed.HostConfig.NetworkMode, "none")
+	}
+	if len(parsed.HostConfig.CapDrop) != 1 || parsed.HostConfig.CapDrop[0] != "ALL" {
+		t.Errorf("HostConfig.CapDrop: got %v, want [ALL]", parsed.HostConfig.CapDrop)
+	}
+
+	// Empty runtime must be omitted entirely (engine picks its default).
+	rawBody = nil
+	exec, err = NewContainerExecutor(ContainerExecutorConfig{
+		Image:      "ubuntu:22.04",
+		HostDir:    "/tmp/workspace",
+		SocketPath: sock,
+	})
+	if err != nil {
+		t.Fatalf("NewContainerExecutor: %v", err)
+	}
+	_ = exec.Close()
+
+	if strings.Contains(string(rawBody), `"Runtime"`) {
+		t.Errorf("create body should omit Runtime when unset; got: %s", string(rawBody))
+	}
+}
+
 func TestNewContainerExecutor_MissingImage(t *testing.T) {
 	_, err := NewContainerExecutor(ContainerExecutorConfig{
 		HostDir: "/tmp/workspace",
--- a/harness/internal/executor/container_test.go
+++ b/harness/internal/executor/container_test.go
@@ -856,6 +856,166 @@ func TestContainerExecutor_Runtime(t *testing.T) {
 	}
 }
 
+// TestContainerExecutor_AllowlistMode_WiringConfig verifies that selecting
+// Network.Mode == "allowlist" causes the create-container request to:
+//
+//   - set NetworkMode to "bridge" (so the container can dial the host),
+//   - inject ExtraHosts so host.docker.internal resolves to the host gateway,
+//   - populate HTTP_PROXY / HTTPS_PROXY / NO_PROXY env in the container.
+//
+// The proxy itself is started on a real local port; we don't exercise it
+// over the wire here (a planted-curl integration test belongs behind a
+// build tag — see container_integration_test.go).
+func TestContainerExecutor_AllowlistMode_WiringConfig(t *testing.T) {
+	var receivedBody containerCreateRequest
+
+	sock, cleanup := mockEngineServer(t, map[string]http.HandlerFunc{
+		"POST /containers/create": func(w http.ResponseWriter, r *http.Request) {
+			_ = json.NewDecoder(r.Body).Decode(&receivedBody)
+			w.Header().Set("Content-Type", "application/json")
+			_ = json.NewEncoder(w).Encode(containerCreateResponse{ID: "test-id"})
+		},
+		"POST /containers/*/start": func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusNoContent)
+		},
+		"POST /containers/*/stop": func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusNoContent)
+		},
+		"DELETE /containers/*": func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusNoContent)
+		},
+	})
+	defer cleanup()
+
+	exec, err := NewContainerExecutor(ContainerExecutorConfig{
+		Image:      "ubuntu:22.04",
+		HostDir:    "/tmp/workspace",
+		SocketPath: sock,
+		Network: &types.NetworkConfig{
+			Mode:      "allowlist",
+			Allowlist: []string{"api.github.com"},
+		},
+	})
+	if err != nil {
+		t.Fatalf("NewContainerExecutor: %v", err)
+	}
+	defer func() { _ = exec.Close() }()
+
+	if receivedBody.HostConfig.NetworkMode != "bridge" {
+		t.Errorf("NetworkMode: got %q, want %q", receivedBody.HostConfig.NetworkMode, "bridge")
+	}
+
+	wantHost := "host.docker.internal:host-gateway"
+	if len(receivedBody.HostConfig.ExtraHosts) != 1 || receivedBody.HostConfig.ExtraHosts[0] != wantHost {
+		t.Errorf("ExtraHosts: got %v, want [%q]", receivedBody.HostConfig.ExtraHosts, wantHost)
+	}
+
+	envSeen := map[string]string{}
+	for _, kv := range receivedBody.Env {
+		parts := strings.SplitN(kv, "=", 2)
+		if len(parts) == 2 {
+			envSeen[parts[0]] = parts[1]
+		}
+	}
+	for _, k := range []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"} {
+		if _, ok := envSeen[k]; !ok {
+			t.Errorf("missing env var %s; got Env=%v", k, receivedBody.Env)
+		}
+	}
+	if !strings.HasPrefix(envSeen["HTTP_PROXY"], "http://host.docker.internal:") {
+		t.Errorf("HTTP_PROXY: got %q, want prefix http://host.docker.internal:<port>", envSeen["HTTP_PROXY"])
+	}
+	if envSeen["HTTP_PROXY"] != envSeen["HTTPS_PROXY"] {
+		t.Errorf("HTTP_PROXY and HTTPS_PROXY should match; got %q vs %q", envSeen["HTTP_PROXY"], envSeen["HTTPS_PROXY"])
+	}
+	if !strings.Contains(envSeen["NO_PROXY"], "127.0.0.1") {
+		t.Errorf("NO_PROXY should at least cover 127.0.0.1; got %q", envSeen["NO_PROXY"])
+	}
+
+	// Hardening defaults should still be present alongside the egress wiring.
+	if len(receivedBody.HostConfig.CapDrop) != 1 || receivedBody.HostConfig.CapDrop[0] != "ALL" {
+		t.Errorf("CapDrop: got %v, want [ALL]", receivedBody.HostConfig.CapDrop)
+	}
+}
+
+// TestContainerExecutor_NoneMode_NoProxyWiring confirms that the proxy
+// lifecycle is only triggered when Network.Mode == "allowlist". For mode
+// "none" the executor must not set ExtraHosts, must not set proxy env,
+// and must keep NetworkMode == "none".
+func TestContainerExecutor_NoneMode_NoProxyWiring(t *testing.T) {
+	var receivedBody containerCreateRequest
+
+	sock, cleanup := mockEngineServer(t, map[string]http.HandlerFunc{
+		"POST /containers/create": func(w http.ResponseWriter, r *http.Request) {
+			_ = json.NewDecoder(r.Body).Decode(&receivedBody)
+			w.Header().Set("Content-Type", "application/json")
+			_ = json.NewEncoder(w).Encode(containerCreateResponse{ID: "test-id"})
+		},
+		"POST /containers/*/start": func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusNoContent)
+		},
+		"POST /containers/*/stop": func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusNoContent)
+		},
+		"DELETE /containers/*": func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusNoContent)
+		},
+	})
+	defer cleanup()
+
+	exec, err := NewContainerExecutor(ContainerExecutorConfig{
+		Image:      "ubuntu:22.04",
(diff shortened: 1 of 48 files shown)


harness/internal/executor/container_test.go
```diff
@@ -782,6 +782,240 @@ func TestContainerExecutor_NetworkModes(t *testing.T) {
...
+		},
+		"DELETE /containers/*": func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusNoContent)
+		},
+	})
+	defer cleanup()
+
+	// Configured runtime must appear in the create-container body.
+	exec, err := NewContainerExecutor(ContainerExecutorConfig{
+		Image:      "ubuntu:22.04",
```
rxbynerd:
```suggestion
		Image:      "ubuntu:24.04",
```

outdated ubuntu version

PR #163, opened 2026-05-10
ci/release: dual-publish container to GHCR + GCAR via Direct WIF

### LINKED ISSUE #159: Publish to Google Cloud Artifact Registry
## Background

[#159 (original ask)] Stirrup currently publishes its container image to GitHub Container Registry only (`ghcr.io/rxbynerd/stirrup`, see `.github/workflows/ci.yml::publish-container`). The forthcoming `K8sExecutor` work will run jobs on the existing `native-runtime` GKE cluster in the `rubynerd-net` GCP project, which has no pull credentials for `ghcr.io`. Granting that cluster a long-lived GitHub PAT just to fix pulls would be a step backwards on the autonomy / Rule-of-Two posture the rest of the harness has been built around (see `docs/safety-rings.md`, `docs/credential-federation.md`).

The fix is to dual-publish the container to both GHCR and Google Cloud Artifact Registry (GCAR), authenticating from GitHub Actions to GCP via Workload Identity Federation — the same keyless pattern stirrup already uses to talk to Anthropic, Vertex AI, and Azure OpenAI from CI (`harness/internal/credential/`). GKE pulls from GCAR with the cluster's existing Google identity; no PATs cross trust boundaries.

GCAR is also a strict superset of GHCR for our purposes: once images live there, we get [Artifact Analysis](https://docs.cloud.google.com/artifact-analysis/docs/artifact-analysis) — automatic continuous vulnerability scanning of every push, plus the ability to attach externally-generated SBOMs via [`gcloud artifacts sbom load`](https://docs.cloud.google.com/artifact-analysis/docs/upload-sboms) so the SPDX + CycloneDX bundles we already produce in `release.yml::sbom` become queryable Container Analysis occurrences instead of inert release-asset blobs. Building a release-gate on top of those occurrences is the obvious follow-up.

## Scope

Three phases, deliverable independently. Phase 1 unblocks `K8sExecutor`; phases 2 and 3 are the security upside.

| Phase | Outcome | Blocking? |
|---|---|---|
| 1. Dual-publish to GHCR + GCAR | `K8sExecutor` can pull from `*-docker.pkg.dev/rubynerd-net/stirrup/stirrup` | Yes — gate for `K8sExecutor` |
| 2. SBOM upload to GCAR | `gcloud artifacts sbom list <image>` shows SPDX+CycloneDX attached to every release tag | No — incremental |
| 3. Artifact Analysis gating + follow-ups | Releases fail (or file issues) on `CRITICAL`/`HIGH` vulnerabilities with fixes available | No — incremental |

## Operator-side bootstrap (one-time, in `rubynerd-net`)

These resources must exist before the workflow changes can succeed. They are non-secret identifiers; once created their full resource paths (project number, pool ID, provider ID) get baked into the workflow as `vars.*` (not secrets — there is nothing sensitive in a WIF resource name and surfacing them in workflow YAML is consistent with how `google-github-actions/auth` documents the pattern).

1. **Enable required APIs:**
   ```sh
   gcloud services enable \
     artifactregistry.googleapis.com \
     containerscanning.googleapis.com \
     containeranalysis.googleapis.com \
     iamcredentials.googleapis.com \
     sts.googleapis.com \
     --project rubynerd-net
   ```
   `containerscanning.googleapis.com` is what activates *automatic* vulnerability scanning on push (per [Artifact Analysis docs](https://docs.cloud.google.com/artifact-analysis/docs/artifact-analysis), scanning is "triggered automatically every time you push a new image" once enabled).

2. **Create the Artifact Registry Docker repository.** Region must be picked to match the `native-runtime` cluster's region for fast pulls; default below to `europe-west4` and revise if the cluster lives elsewhere.
   ```sh
   gcloud artifacts repositories create stirrup \
     --repository-format=docker \
     --location=europe-west4 \
     --description="Stirrup harness container images" \
     --project=rubynerd-net
   ```
   Final image path: `europe-west4-docker.pkg.dev/rubynerd-net/stirrup/stirrup`.

3. **Create the Workload Identity Pool + Provider** for GitHub Actions, scoped to this repository only. Use the [`google-github-actions/auth`](https://github.com/google-github-actions/auth) **Direct Workload Identity Federation** path — no intermediate service account. The provider's attribute condition pins the GitHub repository so a token from another repository cannot exchange for our access token.
   ```sh
   gcloud iam workload-identity-pools create stirrup-gha \
     --location=global --project=rubynerd-net

   gcloud iam workload-identity-pools providers create-oidc stirrup-gha-provider \
     --location=global \
     --workload-identity-pool=gha-pool \
     --issuer-uri="https://token.actions.githubusercontent.com" \
     --attribute-mapping="google.subject=assertion.sub,attribute.repository=assertion.repository,attribute.ref=assertion.ref" \
     --attribute-condition="assertion.repository == 'rxbynerd/stirrup'" \
     --project=rubynerd-net
   ```

4. **Grant the pool's principalSet the IAM roles needed for push and Artifact Analysis writes.** Direct WIF binds roles to the *external identity* (the GHA OIDC subject), not to a service account; this avoids the long-lived service-account-key foot-gun. Required roles (per [GAR push docs](https://docs.cloud.google.com/artifact-registry/docs/docker/authentication) and [SBOM upload docs](https://docs.cloud.google.com/artifact-analysis/docs/upload-sboms)):
   - `roles/artifactregistry.writer` — push container images.
   - `roles/containeranalysis.notes.editor` — create SBOM-reference notes.
   - `roles/containeranalysis.occurrences.editor` — create SBOM-reference occurrences.
   - `roles/containeranalysis.notes.attacher` — attach SBOM-reference notes to images.
   - `roles/storage.objectAdmin` on the Artifact Analysis bucket — `gcloud artifacts sbom load` writes SBOMs to a managed Cloud Storage bucket; without this the load fails after the upload appears to succeed.

   Bind each via `gcloud projects add-iam-policy-binding rubynerd-net --role=<ROLE> --member="principalSet://iam.googleapis.com/projects/<PROJECT_NUMBER>/locations/global/workloadIdentityPools/gha-pool/attribute.repository/rxbynerd/stirrup"`.

5. **Repository variables in GitHub** (Settings → Secrets and variables → Actions → Variables):
   - `GCP_PROJECT_ID = rubynerd-net`
   - `GCP_PROJECT_NUMBER = <numeric>`
   - `GCP_WORKLOAD_IDENTITY_PROVIDER = projects/<NUMBER>/locations/global/workloadIdentityPools/stirrup-gha/providers/stirrup-gha-provider`
   - `GAR_LOCATION = europe-west4` (or wherever the repo was created)
   - `GAR_REPOSITORY = stirrup`
   - `GAR_IMAGE = stirrup` (the image name within the repository)

   These are `vars`, not `secrets`. Resource names are non-sensitive; treating them as such has been a recurring cleanup item across other projects ([cite: hardening guidance in `docs/anthropic-wif.md` follows the same convention]).

### `.github/workflows/ci.yml::publish-container`

Currently the job only pushes to GHCR. Three concrete edits:

1. **Add `id-token: write` permission.** Required for GitHub Actions to mint OIDC tokens for the WIF exchange. Keep `contents: read` and `packages: write`; the latter is still needed for GHCR.
   ```yaml
   permissions:
     contents: read
     id-token: write
     packages: write
   ```

2. **Authenticate to GCP and to GAR** between the existing buildx setup and the `metadata-action` step.
   ```yaml
   - name: Authenticate to Google Cloud (Direct WIF)
     id: gcp-auth
     uses: google-github-actions/auth@v3
     with:
       project_id: ${{ vars.GCP_PROJECT_ID }}
       workload_identity_provider: ${{ vars.GCP_WORKLOAD_IDENTITY_PROVIDER }}
       token_format: access_token

   - name: Log in to Google Artifact Registry
     uses: docker/login-action@v4
     with:
       registry: ${{ vars.GAR_LOCATION }}-docker.pkg.dev
       username: oauth2accesstoken
       password: ${{ steps.gcp-auth.outputs.access_token }}
   ```
   `oauth2accesstoken` is the literal username GAR expects when the password is a short-lived access token (per [docker/login-action](https://github.com/docker/login-action) Google example). The existing GHCR login stays unchanged.

3. **Add the GAR repository to `metadata-action`'s `images:` list.** `docker/build-push-action` will then push every tag to both registries in a single build.
   ```yaml
   - name: Generate image metadata
     id: meta
     uses: docker/metadata-action@v6
     with:
       images: |
         ghcr.io/rxbynerd/stirrup
         ${{ vars.GAR_LOCATION }}-docker.pkg.dev/${{ vars.GCP_PROJECT_ID }}/${{ vars.GAR_REPOSITORY }}/${{ vars.GAR_IMAGE }}
       tags: |
         type=raw,value=latest,enable={{is_default_branch}}
         type=ref,event=branch
         type=sha,prefix=sha-
   ```

4. **Fork-safety guard.** Forked PRs cannot mint id-tokens against `rxbynerd/stirrup`'s WIF provider (the attribute condition pins repository), and the `vars.*` values are not exposed to fork runs. Gate the entire job — already gated on `github.ref == 'refs/heads/main'`, which excludes PRs, but add a belt-and-braces `&& github.repository == 'rxbynerd/stirrup'` so a future change to the trigger condition cannot accidentally surface secret-less WIF auth attempts on a fork's main.

### `.github/workflows/release.yml`

The release pipeline currently produces binaries + source SBOMs but does **not** publish a container at release tag time. Either:

- **Option A (recommended): add a new `publish-container` job in `release.yml`** that mirrors the CI job but tags `vX.Y.Z` (and `latest` for non-prerelease tags) on both registries. Place it after `verify` and before `release`. Same WIF auth steps; `metadata-action` config differs only in the `tags:` block (e.g. `type=semver,pattern={{version}}`, `type=semver,pattern={{major}}.{{minor}}`, `type=raw,value=latest,enable=${{ !contains(env.TAG, '-') }}`).
- **Option B: rely on the `latest`/`sha-*` tags from `ci.yml` and re-tag at release time.** Simpler but creates a window where a tagged commit's container has no semver tag. Reject this option for security-relevant releases.

Pick Option A.

### Multi-architecture build

`Dockerfile` already builds `CGO_ENABLED=0` static binaries via Go's cross-compiler, so multi-arch is essentially free. Add `docker/setup-qemu-action@v3` before `setup-buildx-action`, then pass `platforms: linux/amd64,linux/arm64` to `build-push-action`. The `K8sExecutor` cluster's node pool architecture decides whether arm64 is required at v1 — if `native-runtime` is amd64-only, defer arm64 to a follow-up to keep this issue's blast radius small. Recommended decision: build both and pay the ~30s extra build time, since the binaries are already cross-compiled for release.

### Documentation

- Update `CLAUDE.md`'s "Releases" / "CI" sections so the reader knows the canonical pull paths now include GAR.
- Update `README.md` with `docker pull` examples for both registries.
- New doc `docs/container-publishing.md` covering the GAR bootstrap, the WIF binding model, and how to roll the pool / rotate the provider attribute condition.

### Acceptance criteria — Phase 1

- A push to `main` in `rxbynerd/stirrup` produces an image at `europe-west4-docker.pkg.dev/rubynerd-net/stirrup/stirrup:latest` and `:sha-<7>`, identical digest to `ghcr.io/rxbynerd/stirrup:latest`.
- `gcloud artifacts docker images list europe-west4-docker.pkg.dev/rubynerd-net/stirrup/stirrup` shows the image.
- A `vX.Y.Z` tag push produces matching `:vX.Y.Z` (and `:latest` for non-prereleases) on both registries.
- The `native-runtime` GKE cluster can `docker pull` from GAR using its existing Google identity; no PAT or registry secret is required.
- A PR from a fork still runs `verify`, but `publish-container` is skipped (no WIF attempt, no error).
- `just docker` continues to build the image locally without GAR setup.

### Generate an *image* SBOM, not a source SBOM

The release workflow's existing `sbom` job runs `anchore/sbom-action` against `path: .`, producing an SBOM of the source tree (Go module graph). For Artifact Analysis to map vulnerabilities to actual installed packages, the SBOM must describe the *image* — base layers (`gcr.io/distroless/static-debian12:nonroot`) included. `anchore/sbom-action` accepts `image: <ref>` for this; it shells out to `syft` against the pushed image.

Two sub-options:

- **Sub-option A: regenerate per-image SBOMs after push.** Add a step after `build-push-action` that re-runs `anchore/sbom-action` with `image: <pushed-tag>@<digest>`. Output goes to two new artifact files (`stirrup-<TAG>.image.spdx.json`, `.image.cdx.json`) so the existing release-asset bundle is unaffected.
- **Sub-option B: use `docker/build-push-action`'s built-in `sbom: true`.** This embeds an SBOM as an OCI attestation alongside the image. Useful but *not* equivalent — Artifact Analysis's `sbom load` consumes a JSON file, not an attestation. A pragmatic combination: enable both. `sbom: true` is for downstream consumers using cosign/syft; `sbom load` is for Container Analysis occurrences.

Pick A as primary and B as a cheap addition.

### Upload via `gcloud artifacts sbom load`

After the image is pushed and the SBOM file exists locally:

```yaml
- name: Upload SPDX SBOM to Artifact Analysis
  run: |
    gcloud artifacts sbom load \
      --source stirrup-${TAG}.image.spdx.json \
      --uri ${IMAGE_URI}
- name: Upload CycloneDX SBOM to Artifact Analysis
  run: |
    gcloud artifacts sbom load \
      --source stirrup-${TAG}.image.cdx.json \
      --uri ${IMAGE_URI}
```

`IMAGE_URI` should be the full `<location>-docker.pkg.dev/<project>/<repo>/<image>@sha256:<digest>` form (digest, not tag) — taken from `${{ steps.docker-push.outputs.digest }}`. SPDX 2.2/2.3 and CycloneDX 1.4/1.5 are both supported per [the docs](https://docs.cloud.google.com/artifact-analysis/docs/upload-sboms); `anchore/sbom-action` v0.24.0 emits SPDX 2.3 and CycloneDX 1.5 by default.

### Acceptance criteria — Phase 2

- After every release, `gcloud artifacts sbom list europe-west4-docker.pkg.dev/rubynerd-net/stirrup/stirrup@<digest>` returns two SBOMs (SPDX + CycloneDX).
- Loading either SBOM into a vulnerability scanner (e.g. `grype sbom:<file>`) produces a non-empty package list including the distroless base.
- Source-tree SBOMs (existing release assets) are unchanged.

### Wait for the scan, then query

Automatic scanning is async ("triggered automatically every time you push a new image" but eventual). The release job needs to wait for the scan's `discovery` occurrence to reach `FINISHED_SUCCESS` before reading vulnerability occurrences, otherwise it'll see an empty result set and pass spuriously.

Pattern (bash, in the release workflow):

```sh
IMAGE_URI="europe-west4-docker.pkg.dev/rubynerd-net/stirrup/stirrup@${DIGEST}"
deadline=$(( $(date +%s) + 600 ))
while [ $(date +%s) -lt $deadline ]; do
  state=$(gcloud artifacts docker images list \
    --show-occurrences \
    --occurrence-filter='kind="DISCOVERY"' \
    --format='value(occurrence.discovered.continuousAnalysis)' \
    "${IMAGE_URI}")
  case "$state" in
    *FINISHED*) break ;;
  esac
  sleep 15
done
```

10-minute deadline is conservative; typical scans finish in <2 min. Treat timeout as soft-fail (warn, don't block) so a Google-side outage doesn't wedge releases.

### Severity gate

Per [`gcloud artifacts docker images list --show-occurrences`](https://docs.cloud.google.com/artifact-analysis/docs/os-overview), filter by `kind="VULNERABILITY"` and `--format=json`, then jq for severity and fix availability. Recommended initial threshold: fail the release if any occurrence has `vulnerability.severity == "CRITICAL"` *and* `vulnerability.fixAvailable == true`. Warn on `HIGH` with fix available; allow everything else.

```sh
gcloud artifacts docker images list \
  --show-occurrences \
  --occurrence-filter='kind="VULNERABILITY"' \
  --format=json "${IMAGE_URI}" \
| jq -e '
    [.[] | .occurrence.vulnerability
    | select(.severity == "CRITICAL" and .fixAvailable == true)] | length == 0
  '
```

Decisions an operator should be able to make per release:

- **Block** — exit non-zero, the release fails, a human resolves before re-tagging.
- **Allow with follow-up issue** — for tags already cut where rolling back is high-cost. Use `gh issue create` from the workflow with the vulnerability list as the body.
- **Override** — repo variable like `vars.STIRRUP_RELEASE_VULN_OVERRIDE = vX.Y.Z` lets a specific tag bypass the gate without code changes. Document the override in the issue the workflow files when used.

Recommend starting at "warn + file follow-up" rather than "block" so the gate's calibration is observed before it gates production. Promote to "block" after one quarter of clean signal.

### Acceptance criteria — Phase 3

- A release whose container has a `CRITICAL` vulnerability with `fixAvailable: true` either fails the workflow or files a tracking issue, depending on the chosen mode.
- The gate's outcome is visible in the workflow summary (markdown table of severities × counts) regardless of pass/fail.
- The 10-minute scan-wait timeout warns and continues; it never blocks a release on a Google-side outage.
- An operator can list current vulnerabilities for any released digest with one `gcloud` command, captured in the new docs.

## Risks and known unknowns

1. **GitHub Actions OIDC issuer + tag pinning.** If GitHub ever rotates `token.actions.githubusercontent.com`'s issuer URL or JWKS keys, the WIF provider must be updated. Low probability, but the operator-side bootstrap doc should call out the issuer URL as a thing to monitor. `google-github-actions/auth` itself handles JWKS caching transparently.
2. **Provider attribute condition is the *only* repo guard.** A misconfigured condition (e.g. omitting it, or matching `assertion.repository_owner` instead of `assertion.repository`) lets any repo in the org exchange tokens. Bootstrap script must include the condition; verify with `gcloud iam workload-identity-pools providers describe` after creation.
3. **Forks and PRs from forks.** Auth steps must be gated so forked PRs don't attempt WIF auth (which would silently fail because the provider doesn't trust them). Existing `if: github.ref == 'refs/heads/main'` on `publish-container` already handles this for CI; the new release-time job inherits the tag-trigger guard. Document that running the workflow from a fork is unsupported by design.
4. **GAR repo region mismatch with `native-runtime` cluster.** Cross-region pulls work but are slower and incur egress. Pick the region the cluster lives in. Verify before bootstrap.
5. **`gcloud artifacts sbom load` requires Cloud Storage permissions** that aren't obvious from the IAM role names. The docs call out `roles/storage.objectAdmin` on the managed bucket for "default Cloud Storage bucket management." Without it, the load command exits 0 but the SBOM never appears in `gcloud artifacts sbom list`. Bootstrap must include this binding; smoke-test by listing SBOMs after the first push.
6. **SBOM size and format drift.** GCAR accepts SPDX 2.2/2.3 and CycloneDX 1.4/1.5. `anchore/sbom-action` versions can drift the emitted format minor without warning; pin the action SHA the same way `release.yml` already pins (`anchore/sbom-action@e22c389...`).
7. **Distroless base + Go binary leaves a small SBOM.** Most vulnerabilities will be in the base image's libc/ssl, not in our Go deps. That's fine — Artifact Analysis uses [OS package scanning](https://docs.cloud.google.com/artifact-analysis/docs/os-overview) for those — but expect the gate's signal-to-noise ratio to depend mostly on how often Google patches distroless rather than on our own dependency hygiene.
8. **Continuous Analysis cost.** GCAR scanning is billed per scanned image. At Stirrup's release cadence this is negligible; flag it for awareness.
9. **`google-github-actions/auth` v3 vs v2.** The action upgraded from v2→v3 with breaking changes around input names (`workload_identity_provider` → `workload_identity_provider`, plus `service_account` no longer required for direct WIF). Use v3 and pin to a SHA, mirroring how `release.yml` already SHA-pins `anchore/sbom-action` and `softprops/action-gh-release`.
10. **`docker/login-action` username quirk.** GAR demands the literal string `oauth2accesstoken` as the username; pasting a service-account email in there fails opaquely. Capture this in a comment on the login step.

## Out of scope

- Migrating *binaries* (the `release.yml` artifact bundle) to GAR. Generic artifacts in GAR are possible but GitHub Releases remain the canonical binary distribution path.
- cosign / Sigstore signing of GAR images. The release workflow already has a commented-out cosign seam for binaries; container signing will follow the same staged rollout in a separate issue.
- Building images for additional platforms beyond `linux/amd64` and `linux/arm64`.
- BinAuthz attestations. Once Phase 3 lands, BinAuthz can consume the same vulnerability occurrences to gate `kubectl apply`; that's a follow-up scoped to the K8sExecutor cluster setup.
- Mirroring images *into* GHCR from GAR (the inverse direction). Not needed.

## Implementation step order

1. Operator bootstrap (Phase 1 prerequisites): APIs enabled, GAR repo created, WIF pool/provider created, IAM bound, repo `vars.*` populated. Smoke-test by running `gcloud auth print-access-token --impersonate-service-account ...` from a throwaway workflow to confirm the WIF exchange works *before* touching `ci.yml`.
2. Edit `.github/workflows/ci.yml::publish-container`: add `id-token: write`, add `google-github-actions/auth@v3` step (SHA-pinned), add `docker/login-action@v4` step for GAR, extend `metadata-action`'s `images:` list, add `docker/setup-qemu-action@v3` and `platforms: linux/amd64,linux/arm64`. Verify on a personal fork's `main` push first by temporarily relaxing the repo-pin guard (then revert).
3. Add `publish-container` job to `.github/workflows/release.yml` mirroring (2) but with semver tagging. Include the `digest` output capture for Phase 2.
4. Update `Dockerfile` if anything blocks multi-arch (likely nothing — the build is `CGO_ENABLED=0`).
5. Documentation: `docs/container-publishing.md`, `CLAUDE.md` references, `README.md` pull examples. Phase 1 ships behind these.
6. Phase 2: regenerate SBOMs from the pushed image (anchore/sbom-action with `image:` and `@sha256:` digest); add `gcloud artifacts sbom load` steps for SPDX and CycloneDX; verify with `gcloud artifacts sbom list`.
7. Phase 3: discovery-state poll loop; vulnerability query + jq filter; choose initial mode (warn + follow-up issue); wire up `gh issue create` with a templated body; document the override variable.
8. Promote Phase 3 mode from "warn" to "block" after a calibration window — separate small PR.

## References

- Issue text and links from #159 (Artifact Analysis, SBOM upload).
- `google-github-actions/auth`: <https://github.com/google-github-actions/auth>
- `docker/login-action` Google example: <https://github.com/docker/login-action#google-artifact-registry-gar>
- Artifact Analysis overview: <https://docs.cloud.google.com/artifact-analysis/docs/artifact-analysis>
- Upload SBOMs: <https://docs.cloud.google.com/artifact-analysis/docs/upload-sboms>
- OS scanning (vulnerability query syntax): <https://docs.cloud.google.com/artifact-analysis/docs/os-overview>
- Existing stirrup credential federation patterns: `harness/internal/credential/`, `docs/credential-federation.md`.
- Existing CI/release workflows: `.github/workflows/ci.yml`, `.github/workflows/release.yml`, `.github/workflows/_verify.yml`.
- `Dockerfile` (build target).

rxbynerd:
## Summary

Phase 1 of #159: dual-publish the stirrup container to **GitHub Container Registry** (existing) **and Google Cloud Artifact Registry** (new). Authenticates from GitHub Actions to GCP via Direct Workload Identity Federation — no static keys, no service-account key, no PAT crosses trust boundaries. Multi-arch (`linux/amd64` + `linux/arm64`) from day one.

Phases 2 (image-SBOM upload to Artifact Analysis) and 3 (vulnerability gating on release) are intentionally **deferred to follow-up PRs**. This PR is the dependency unblocker for `K8sExecutor` (which will pull from the `native-runtime` GKE cluster's own Google identity, no GHCR PAT required).

Closes #159.

## What changed

- **`.github/workflows/ci.yml::publish-container`** — adds `id-token: write`, fork-safety repo pin (`github.repository == 'rxbynerd/stirrup'`), QEMU emulation, `google-github-actions/auth@v3` (SHA-pinned), GAR login via `oauth2accesstoken`, dual-registry image list in `metadata-action`, multi-arch `platforms: linux/amd64,linux/arm64`. `VERSION` build-arg parameterised via a step output rather than hard-coded `main`.
- **`.github/workflows/release.yml`** — new `publish-container` job parallel to `build-binaries` / `sbom` / `changelog`. Same WIF auth steps; semver tag rules (`{{version}}`, `{{major}}.{{minor}}`, `latest` for non-prereleases). The `release` job now `needs: publish-container`. Header pipeline diagram updated to reflect fan-out DAG.
- **All four docker/* actions in both publish-container jobs are SHA-pinned** (`docker/login-action`, `docker/setup-buildx-action`, `docker/build-push-action`, `docker/metadata-action`) — the auth action that mints the WIF token is pinned with a "non-negotiable" rationale; pinning the action that *consumes* the token is the same rationale.
- **`docs/container-publishing.md`** — new operator walkthrough mirroring `docs/anthropic-wif.md` / `docs/azure-workload-identity.md`. Covers GCP bootstrap, WIF binding model, provider-condition rotation, verification, and risks. Phase 1 vs Phase 2 IAM bindings are split into separate sections so operators following the doc today bind only what Phase 1 needs.
- **`CLAUDE.md`** and **`README.md`** — `### CI` / `### Releases` / `### Container image` references updated to call out both pull paths.

## Operator-side work done in this session

To verify the workflow end-to-end I (with your authorization in the original ask) ran the full operator-side bootstrap in project `rubynerd-net`:

- APIs enabled: `artifactregistry`, `containerscanning`, `containeranalysis`, `iamcredentials`, `sts`
- Artifact Registry repo `stirrup` at `europe-west4` (DOCKER format)
- WIF pool `stirrup-gha` + provider `stirrup-gha-provider`
- **Provider attribute condition is locked**: `assertion.repository == 'rxbynerd/stirrup' && (assertion.ref == 'refs/heads/main' || assertion.ref.startsWith('refs/tags/v'))` — pushes the trust decision from YAML (any collaborator can edit) down to the GCP credential layer
- Five IAM roles bound to the principalSet: `artifactregistry.writer` (Phase 1), plus `containeranalysis.notes.editor` / `.occurrences.editor` / `.notes.attacher` / `storage.objectAdmin` (Phase 2 prereqs, bound now to avoid a second bootstrap)

### Action required from you before this works in CI

Set six repository variables at https://github.com/rxbynerd/stirrup/settings/variables/actions:

| Name | Value |
|---|---|
| `GCP_PROJECT_ID` | `rubynerd-net` |
| `GCP_PROJECT_NUMBER` | `163317929648` |
| `GCP_WORKLOAD_IDENTITY_PROVIDER` | `projects/163317929648/locations/global/workloadIdentityPools/stirrup-gha/providers/stirrup-gha-provider` |
| `GAR_LOCATION` | `europe-west4` |
| `GAR_REPOSITORY` | `stirrup` |
| `GAR_IMAGE` | `stirrup` |

Until these are set, the first `Authenticate to Google Cloud` step on a `main` push will fail loudly (which is the intended behaviour — fail-closed). The existing GHCR publish path is unaffected.

## Review cycle

Five reviewers ran in parallel (code, security, spec-compliance, test-coverage, api-design). Findings were consolidated by a synthesizer into a remediation brief, and a remediation commit set landed before this PR was opened. Highlights:

- **2 MUST-FIX both applied**: SHA-pin the four docker/* actions in publish-container; tighten the WIF provider attribute-condition to lock issuance to `main`+`v*` (the live provider config was updated, and `docs/container-publishing.md` now documents the locked condition).
- **4 SHOULD-FIX applied**: explain the deliberate `event_name` omission on the release-time job; split Phase 1 vs Phase 2 IAM bindings in the doc; update the release.yml header pipeline diagram; parameterise the `VERSION` build-arg in ci.yml.
- **10 follow-ups deferred** to subsequent issues (actionlint, smoke-gar-publish workflow, narrow storage.objectAdmin to bucket scope, GHA Environment with reviewers, reusable `_publish-container.yml`, pre-existing softprops SHA-pin comment mismatch in main, GAR_IMAGE_* multi-image naming, arm64 functional smoke, release recovery doc, GCP_PROJECT_NUMBER var removal). All are tracked in the persisted memory file for future-session pickup of Phases 2/3.

## Phase 2 and 3 status

Out of scope for this PR. Phase 2 (image-target SBOM emission + `gcloud artifacts sbom load`) will land separately; the Phase 2 IAM roles are already bound so it can be a YAML-only change. Phase 3 (vulnerability gating on release tags, calibrated as "warn + file follow-up issue" before promoting to "block") follows Phase 2. Notes preserved for that work.

## Test plan

- [ ] Confirm CI publish path: once the six `vars.*` above are set, the next push to `main` should produce identical digests at `ghcr.io/rxbynerd/stirrup:latest` and `europe-west4-docker.pkg.dev/rubynerd-net/stirrup/stirrup:latest`.
- [ ] `gcloud artifacts docker images list europe-west4-docker.pkg.dev/rubynerd-net/stirrup/stirrup` shows the image after first publish.
- [ ] On a `v*.*.*` tag push: matching `:vX.Y.Z`, `:X.Y`, and `:latest` (for non-prereleases) on both registries.
- [ ] Confirm `docker manifest inspect europe-west4-docker.pkg.dev/rubynerd-net/stirrup/stirrup:<tag>` shows entries for both `linux/amd64` and `linux/arm64`.
- [ ] Verify a workflow run from any branch other than `main` (with `id-token: write`) fails the WIF exchange at the GCP layer (HTTP 403), not at the YAML guard.
- [ ] `just docker` continues to build locally without any GAR setup or `gcloud` dependency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,9 +128,21 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v6
 
-      - name: Compute short SHA
+      # Mirrors the eval-gate version-computation step. The `if:` guard
+      # above currently restricts this job to refs/heads/main pushes, so
+      # `label` will always be `main` today; computing it indirectly
+      # keeps the build-args block honest if that condition is ever
+      # relaxed (e.g. tag-triggered runs added) and prevents the
+      # VERSION literal from drifting from the actual ref.
+      - name: Compute version label
         id: ver
-        run: echo "sha=${GITHUB_SHA::7}" >> "$GITHUB_OUTPUT"
+        run: |
+          if [ "${{ github.ref }}" = "refs/heads/main" ]; then
+            echo "label=main" >> "$GITHUB_OUTPUT"
+          else
+            echo "label=dev" >> "$GITHUB_OUTPUT"
+          fi
+          echo "sha=${GITHUB_SHA::7}" >> "$GITHUB_OUTPUT"
 
       # SHA-pinned to v3.7.0. SHA verified via the GitHub API on
       # 2026-05-10 (`gh api repos/docker/setup-qemu-action/git/refs/tags/v3.7.0`,
@@ -219,5 +231,5 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           build-args: |
-            VERSION=main
+            VERSION=${{ steps.ver.outputs.label }}
             COMMIT=${{ steps.ver.outputs.sha }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,11 +138,23 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3.7.0
 
+      # SHA-pinned to v4.0.0. SHA verified via the GitHub API on
+      # 2026-05-10 (`gh api repos/docker/setup-buildx-action/git/refs/tags/v4.0.0`,
+      # which also matches the v4 floating tag's current target). Pinned
+      # for the same reason as gcp-auth: this action runs inside the
+      # job that holds a live GAR access token; a retargeted floating
+      # tag could exfiltrate it.
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v4
+        uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
 
+      # SHA-pinned to v4.1.0. SHA verified via the GitHub API on
+      # 2026-05-10 (`gh api repos/docker/login-action/git/refs/tags/v4.1.0`,
+      # which also matches the v4 floating tag's current target). This
+      # action receives the GHCR/GAR credentials as `password:` —
+      # downstream consumer of the token gcp-auth mints; pinning to an
+      # immutable commit is non-negotiable here too.
       - name: Log in to GitHub Container Registry
-        uses: docker/login-action@v4
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
@@ -163,7 +175,7 @@ jobs:
           token_format: access_token
 
       - name: Log in to Google Artifact Registry
-        uses: docker/login-action@v4
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
         with:
           registry: ${{ vars.GAR_LOCATION }}-docker.pkg.dev
           # GAR demands the literal string "oauth2accesstoken" as the
@@ -173,9 +185,15 @@ jobs:
           username: oauth2accesstoken
           password: ${{ steps.gcp-auth.outputs.access_token }}
 
+      # SHA-pinned to v6.0.0. SHA verified via the GitHub API on
+      # 2026-05-10 (`gh api repos/docker/metadata-action/git/refs/tags/v6.0.0`,
+      # which also matches the v6 floating tag's current target). Pinned
+      # for the same reason as login-action: runs inside the credentialled
+      # publish job and constructs the tag list that build-push-action
+      # consumes — a retargeted tag could redirect pushes.
       - name: Generate image metadata
         id: meta
-        uses: docker/metadata-action@v6
+        uses: docker/metadata-action@030e881283bb7a6894de51c315a6bfe6a94e05cf # v6.0.0
         with:
           images: |
             ghcr.io/rxbynerd/stirrup
@@ -185,8 +203,14 @@ jobs:
             type=ref,event=branch
             type=sha,prefix=sha-
 
+      # SHA-pinned to v7.1.0. SHA verified via the GitHub API on
+      # 2026-05-10 (`gh api repos/docker/build-push-action/git/refs/tags/v7.1.0`,
+      # which also matches the v7 floating tag's current target). This
+      # action performs the actual push using the live GHCR + GAR
+      # credentials; pinning to an immutable commit is non-negotiable
+      # here for the same reason as gcp-auth and login-action.
       - name: Build and push container image
-        uses: docker/build-push-action@v7
+        uses: docker/build-push-action@bcafcacb16a39f128d818304e6c9c0c18556b85f # v7.1.0
         with:
           context: .
           file: ./Dockerfile
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,9 +114,15 @@ jobs:
     name: Publish Container
     runs-on: ubuntu-latest
     needs: verify
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    # Belt-and-braces fork guard: the WIF provider's attribute condition
+    # already pins assertion.repository == 'rxbynerd/stirrup', but adding
+    # the repository check here means a future change to the trigger
+    # condition (e.g. running on forked main branches) cannot accidentally
+    # surface a silent WIF auth failure on someone else's fork.
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main' && github.repository == 'rxbynerd/stirrup'
     permissions:
       contents: read
+      id-token: write
       packages: write
     steps:
       - name: Check out repository
@@ -125,6 +131,12 @@ jobs:
       - name: Compute short SHA
         id: ver
         run: echo "sha=${GITHUB_SHA::7}" >> "$GITHUB_OUTPUT"
+
+      # SHA-pinned to v3.7.0. SHA verified via the GitHub API on
+      # 2026-05-10 (`gh api repos/docker/setup-qemu-action/git/refs/tags/v3.7.0`,
+      # which also matches the v3 floating tag's current target).
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3.7.0
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
@@ -136,11 +148,38 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      # SHA-pinned to v3.0.0. SHA verified via the GitHub API on
+      # 2026-05-10 (`gh api repos/google-github-actions/auth/git/refs/tags/v3.0.0`,
+      # which also matches the v3 floating tag's current target). This
+      # action mints a Google access token from the GHA OIDC JWT and is
+      # the trust hinge for GAR push — pinning to an immutable commit is
+      # non-negotiable.
+      - name: Authenticate to Google Cloud (Direct WIF)
+        id: gcp-auth
+        uses: google-github-actions/auth@7c6bc770dae815cd3e89ee6cdf493a5fab2cc093 # v3.0.0
+        with:
+          project_id: ${{ vars.GCP_PROJECT_ID }}
+          workload_identity_provider: ${{ vars.GCP_WORKLOAD_IDENTITY_PROVIDER }}
+          token_format: access_token
+
+      - name: Log in to Google Artifact Registry
+        uses: docker/login-action@v4
+        with:
+          registry: ${{ vars.GAR_LOCATION }}-docker.pkg.dev
+          # GAR demands the literal string "oauth2accesstoken" as the
+          # username when the password is a short-lived access token.
+          # Substituting a service-account email here fails opaquely —
+          # do not "fix" this to something more meaningful.
+          username: oauth2accesstoken
+          password: ${{ steps.gcp-auth.outputs.access_token }}
+
       - name: Generate image metadata
         id: meta
         uses: docker/metadata-action@v6
         with:
-          images: ghcr.io/rxbynerd/stirrup
+          images: |
+            ghcr.io/rxbynerd/stirrup
+            ${{ vars.GAR_LOCATION }}-docker.pkg.dev/${{ vars.GCP_PROJECT_ID }}/${{ vars.GAR_REPOSITORY }}/${{ vars.GAR_IMAGE }}
           tags: |
             type=raw,value=latest,enable={{is_default_branch}}
             type=ref,event=branch
@@ -152,6 +191,7 @@ jobs:
           context: .
           file: ./Dockerfile
           push: true
+          platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           build-args: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -284,11 +284,23 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3.7.0
 
+      # SHA-pinned to v4.0.0. SHA verified via the GitHub API on
+      # 2026-05-10 (`gh api repos/docker/setup-buildx-action/git/refs/tags/v4.0.0`,
+      # which also matches the v4 floating tag's current target). Pinned
+      # for the same reason as gcp-auth: this action runs inside the
+      # job that holds a live GAR access token; a retargeted floating
+      # tag could exfiltrate it.
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v4
+        uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
 
+      # SHA-pinned to v4.1.0. SHA verified via the GitHub API on
+      # 2026-05-10 (`gh api repos/docker/login-action/git/refs/tags/v4.1.0`,
+      # which also matches the v4 floating tag's current target). This
+      # action receives the GHCR/GAR credentials as `password:` —
+      # downstream consumer of the token gcp-auth mints; pinning to an
+      # immutable commit is non-negotiable here too.
       - name: Log in to GitHub Container Registry
-        uses: docker/login-action@v4
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
@@ -309,7 +321,7 @@ jobs:
           token_format: access_token
 
       - name: Log in to Google Artifact Registry
-        uses: docker/login-action@v4
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
         with:
           registry: ${{ vars.GAR_LOCATION }}-docker.pkg.dev
           # GAR demands the literal string "oauth2accesstoken" as the
@@ -319,9 +331,15 @@ jobs:
           username: oauth2accesstoken
           password: ${{ steps.gcp-auth.outputs.access_token }}
 
+      # SHA-pinned to v6.0.0. SHA verified via the GitHub API on
+      # 2026-05-10 (`gh api repos/docker/metadata-action/git/refs/tags/v6.0.0`,
+      # which also matches the v6 floating tag's current target). Pinned
+      # for the same reason as login-action: runs inside the credentialled
+      # publish job and constructs the tag list that build-push-action
+      # consumes — a retargeted tag could redirect pushes.
       - name: Generate image metadata
         id: meta
-        uses: docker/metadata-action@v6
+        uses: docker/metadata-action@030e881283bb7a6894de51c315a6bfe6a94e05cf # v6.0.0
         with:
           images: |
             ghcr.io/rxbynerd/stirrup
@@ -335,8 +353,14 @@ jobs:
             type=semver,pattern={{major}}.{{minor}}
             type=raw,value=latest,enable=${{ !contains(needs.validate-tag.outputs.tag, '-') }}
 
+      # SHA-pinned to v7.1.0. SHA verified via the GitHub API on
+      # 2026-05-10 (`gh api repos/docker/build-push-action/git/refs/tags/v7.1.0`,
+      # which also matches the v7 floating tag's current target). This
+      # action performs the actual push using the live GHCR + GAR
+      # credentials; pinning to an immutable commit is non-negotiable
+      # here for the same reason as gcp-auth and login-action.
       - name: Build and push container image
-        uses: docker/build-push-action@v7
+        uses: docker/build-push-action@bcafcacb16a39f128d818304e6c9c0c18556b85f # v7.1.0
         with:
           context: .
           file: ./Dockerfile
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -259,6 +259,18 @@ jobs:
     # mirror the belt-and-braces guard from ci.yml::publish-container so
     # a future change to the trigger condition cannot accidentally
     # surface silent WIF auth failures on someone else's fork.
+    #
+    # `github.event_name` is intentionally NOT part of this guard: the
+    # job must fire on both `push` (tag push from main) and
+    # `workflow_dispatch` (operator re-run against an existing tag when
+    # a transient GAR/GHCR failure left the release incomplete). Adding
+    # an event_name check would silently break the dispatch recovery
+    # path. The enforcement layers that DO matter are the repository
+    # guard above and the WIF provider's attribute-condition (see
+    # docs/container-publishing.md), which pins both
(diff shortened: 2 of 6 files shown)


rxbynerd:
Three additional commits on top of the original review pass:

- `84befdf` — `ci: add smoke-gar-publish.yml for manual WIF + GAR auth dry-run`
- `4f7c2bb` — `docs: drop GCP_PROJECT_NUMBER from repo vars; only needed locally during bootstrap`
- `2ac0548` — `release: reconcile softprops/action-gh-release prose comment with pinned v3.0.0 SHA`

### Updated verification flow (pre-merge)

The `vars.*` table is now **five** values, not six (`GCP_PROJECT_NUMBER` removed — no workflow consumed it):

| Name | Value |
|---|---|
| `GCP_PROJECT_ID` | `rubynerd-net` |
| `GCP_WORKLOAD_IDENTITY_PROVIDER` | `projects/163317929648/locations/global/workloadIdentityPools/stirrup-gha/providers/stirrup-gha-provider` |
| `GAR_LOCATION` | `europe-west4` |
| `GAR_REPOSITORY` | `stirrup` |
| `GAR_IMAGE` | `stirrup` |

Once those are populated, you can exercise the WIF exchange **before** merging:

```sh
gh workflow run smoke-gar-publish.yml --ref gh-159
gh run watch
```

The smoke workflow auths to GCP via the same SHA-pinned `google-github-actions/auth` action used by the real publish path, mints a federated access token, and runs `gcloud artifacts repositories describe stirrup` to confirm the IAM binding works. It does not push anything. Expected runtime ~30 seconds. If it passes, the production publish path should work on the first `main` push after merge.

If you'd rather just merge and watch the next push to main, that's still fine — the failure mode is fail-closed and existing GHCR publishing is unaffected.

Issue #172, opened 2026-05-11
provider/openai-responses: empty tool_result content omits required `output` field, breaks Responses API mid-run


rxbynerd:
## Summary

The OpenAI Responses API provider adapter (`harness/internal/provider/openai_responses.go`) drops the `output` field from `function_call_output` items when the corresponding stirrup `tool_result` has an empty `Content` string. The OpenAI Responses API requires `output` to be present on every `function_call_output` item and rejects requests that omit it with HTTP 400. This crashes any run where a tool legitimately produces empty output (e.g. a `run_command` whose command writes nothing to stdout).

In the wild, this manifests as a mid-run error:

```
provider stream failed ... openai responses API returned status 400:
  Missing required parameter: 'input[13].output'.
```

The failure is silent until the next turn — the harness happily records an empty tool result, sends it to the model, and only finds out it was malformed when OpenAI rejects the next request.

## Reproduction

```sh
OPENAI_KEY="..." ./stirrup harness \\
  --api-key-ref secret://OPENAI_KEY \\
  --provider openai-responses \\
  --model gpt-5.4-nano \\
  --prompt "Review the last commit for factual inaccuracies"
```

What happens in this run (see captured log in description below):

1. Turns 0–3 succeed normally (model issues `list_directory`, `run_command`, `read_file` calls).
2. Turn 4 dispatches a `run_command` whose stdout is empty. The transport emits:
   ```json
   {"type":"tool_result","toolUseId":"call_tf5TE1X5MpOPq8XmUjEf1QPd"}
   ```
   (notice the missing `content` key — `omitempty` on the wire-level type drops it).
3. The harness appends a `tool_result` content block with `Content: ""` to the message history.
4. On turn 5, `translateMessagesResponses` emits a `function_call_output` input item with `Output: ""`. Because the struct field has `json:"output,omitempty"`, the marshalled JSON omits the `output` key entirely.
5. OpenAI rejects the request:
   ```
   provider stream failed ... openai responses API returned status 400:
     Missing required parameter: 'input[13].output'.
   ```
   Run terminates with outcome `error`.

Reproducing without an API key: any tool call where `block.Content == ""` triggers the bug. A unit test (see below) is sufficient.

## Root Cause

`harness/internal/provider/openai_responses.go`, lines 114–122:

```go
type responsesInput struct {
    Type      string                  `json:"type"`                // "message" | "function_call" | "function_call_output"
    Role      string                  `json:"role,omitempty"`      // for "message"
[...]

PR #181, opened 2026-05-12
eval: wire full RunConfig surface through suite → task → harness invocation

### LINKED ISSUE #177: eval: wire full RunConfig surface through suite → task → harness invocation
## Problem

`stirrup-eval` currently exposes only a slice of the harness's configuration surface. The runner shells out to `stirrup harness` with exactly five flags ([`eval/runner/runner.go:261-268`](https://github.com/rxbynerd/stirrup/blob/main/eval/runner/runner.go#L261-L268)):

```go
args := []string{
    "harness",
    "--prompt", task.Prompt,
    "--mode", taskMode(task),
    "--workspace", workspaceDir,
    "--trace", traceFile,
    "--timeout", "300",
}
```

Everything else — provider type, model, API key reference, credential federation, permission policy, executor, edit strategy, verifier, guardrail, code scanner, observability, max-turns, transport — must come from the harness binary's defaults or the operator's external environment. The eval framework cannot express any of these per-task or per-suite.

This surfaced concretely while writing the regression suite for #172 (the openai-responses empty `output` bug). The suite at [`eval/suites/openai-responses-empty-tool-output.hcl`](eval/suites/openai-responses-empty-tool-output.hcl) (added in #176) cannot, on its own, ensure the harness binary it shells out to is actually configured for the `openai-responses` provider. The suite file *describes* a regression scenario but cannot *enforce* the configuration that makes the scenario meaningful — that responsibility leaks to whoever invokes `stirrup-eval run`.

More broadly: if `stirrup-eval` can only cover a subset of the harness's feature flags, then it cannot regression-test or characterise the rest of stirrup's surface. That includes any provider-specific wire-format behaviour (the #172 bug class generalises to every provider adapter), any credential-federation path (`anthropic-wif`, `gcp-workload-identity-federation`, `azure-workload-identity`, `web-identity`), the container executor, the egress proxy, Cedar permission policies, all five safety rings, all four read-only modes, the GuardRail stages, the code scanner variants, and the OTel/lakehouse pipeline. None of these have an eval-driven characterisation today, and the gap is structural: the runner can't pass the configuration to the harness.

## Stirrup-eval has two audiences and this gap hurts both

1. **Internal development** — bug regressions like #172, multi-provider correctness suites, capability matrices, performance characterisation under different executor / safety-ring settings.
2. **Stirrup's users** — operators tuning their own deployments. An operator running stirrup in a production posture (Azure OpenAI + WIF + container executor + Cedar policy + GuardRail) cannot use `stirrup-eval` to regression-test their actual configuration. The eval framework gives them a stripped-down toy harness, not the one they're running.

A complete eval surface would let users author suites that mirror their deployment posture and run them as part of their own CI / pre-release gates.

## The abstractions already exist

This isn't a greenfield design — most of the type surface is already in place:

- **`stirrup harness --config <runconfig.json>`** ([`harness/cmd/stirrup/cmd/harness.go:497`](https://github.com/rxbynerd/stirrup/blob/main/harness/cmd/stirrup/cmd/harness.go#L497)) already accepts a full RunConfig JSON file. The flag's docstring explicitly documents the precedence rule: \"Explicit flags still override individual fields; unset flags do not.\" This is the wire we want.
- **`types.RunConfigOverrides`** ([`types/eval.go:51-59`](https://github.com/rxbynerd/stirrup/blob/main/types/eval.go#L51-L59)) already exists and is the documented per-variant override carrier for the `Experiment` type. The fields are sparse on purpose (`Provider`, `ModelRouter`, `ContextStrategy`, `EditStrategy`, `Verifier`, `MaxTurns`) — they're the components stirrup considers experiment-worthy. The same shape is appropriate for per-task suite overrides.
- **`RunConfig.Redact()`** strips secret references before any trace or recording is persisted. This invariant must hold across this change; the eval runner must not log a config that contains resolved secrets, only redacted ones.

The gap is purely the wiring between these existing pieces.

### HCL surface — suite-level and per-task

```hcl
suite \"openai-responses-empty-tool-output-regression\" {
  description = \"...\"

  # Suite-level baseline: applies to every task unless overridden.
  run_config_file = \"eval/configs/openai-responses-base.json\"  # optional

  # OR inline. Mutually exclusive with run_config_file.
  run_config {
    provider {
      type        = \"openai-responses\"
      api_key_ref = \"secret://OPENAI_KEY\"
    }
    model_router {
      model = \"gpt-5.4-nano\"
    }
    max_turns = 10
  }

  task \"empty-stdout-run-command-completes\" {
    description = \"...\"
    prompt      = \"...\"

    # Per-task override layer. Sparse — only the fields you want to change.
    run_config_overrides {
      max_turns = 4
    }

    judge { ... }
  }
}
```

Two patterns to support:
- **File reference** (`run_config_file = \"path.json\"`) — points to a checked-in or generated RunConfig JSON. Best for sharing posture across many suites and for matching what `stirrup harness --config` already consumes.
- **Inline HCL block** (`run_config { ... }`) — best for one-off suites where the config is small and the suite is self-contained.

Per-task `run_config_overrides` is sparse (`RunConfigOverrides` shape) and applied on top of the suite-level baseline.

### Runner behaviour

For each task, the runner:
1. Loads the suite-level baseline (file or inline block) into a `RunConfig`.
2. Applies the task's `RunConfigOverrides` on top.
3. Writes the resulting `RunConfig` to a per-task temp file inside the task's tmpdir (alongside the existing `trace.jsonl` artifact retention).
4. Invokes `stirrup harness --config <temp-runconfig.json> --prompt ... --workspace ... --trace ... --timeout ...`.
5. Honours the documented `--config` precedence: explicit flags still override (so `--workspace`, `--trace`, `--timeout` stay authoritative on the runner's side).

The temp file is removed with the rest of the workspace via the existing `defer os.RemoveAll(tmpDir)` ([`eval/runner/runner.go:250`](https://github.com/rxbynerd/stirrup/blob/main/eval/runner/runner.go#L250)). The retained artifact (under `--output`) should include the redacted RunConfig for reproducibility — call `RunConfig.Redact()` before writing.

### CLI / dry-run / compare

- `--dry-run` should validate the config: parse it, run `ValidateRunConfig`, fail fast with a clear error if the suite's config is invalid for the requested mode (e.g. read-only modes with `write_file` in the tool list).
- `compare` does not need to change; baseline diffs are still by `(suite_id, task_id)`.

### Backwards compatibility

Existing suites must continue to work unchanged. A suite with no `run_config_file`, no `run_config` block, and no per-task `run_config_overrides` must behave exactly as it does today — the runner falls back to invoking the harness with the current five-flag set. The new fields are purely additive.

## Out of scope (track separately if pursued)

- **Per-task provider override without a full config.** The `RunConfigOverrides` shape already supports this (`Provider *ProviderConfig`), so it falls out of the above naturally. No separate work.
- **Experiment runner integration.** `types.Experiment` and `types.ExperimentVariant` already use `RunConfigOverrides`. Once the suite/task path consumes this shape, plumbing through to `Experiment` should be a small follow-up rather than its own design effort.
- **Replay-mode interaction.** `ReplayProvider` sidesteps HTTP marshalling, so a suite that overrides `Provider` but is run in replay mode would have a configuration with no effect. Document this clearly in the eval reference and consider warning at suite-load time.
- **Multi-config matrix runs.** \"Run this suite under N different RunConfigs\" is a useful capability for cross-provider correctness suites (the #172 regression generalises across providers). This may be a separate flag on `stirrup-eval run` (e.g. `--config-matrix dir/*.json`) rather than something baked into the suite file. Tracking as future work.
- **Sensitive-data redaction at eval time.** The retained config artifact needs `Redact()`; if the user's RunConfig file contains a `secret://` reference, that reference is fine, but if a future addition allowed a resolved secret value in a config file, redaction must still hold. Cross-reference the existing invariant in `CLAUDE.md`.

## Acceptance criteria

- [ ] `eval/spec/hcl.go` parses optional `run_config_file`, `run_config { ... }`, and per-task `run_config_overrides { ... }` blocks.
- [ ] `eval/spec/hcl.go` rejects suites that set both `run_config_file` and `run_config` (mutually exclusive).
- [ ] `types.EvalSuite` and `types.EvalTask` carry the new fields (or a new `RunConfigSource` wrapper).
- [ ] `eval/runner/runner.go` resolves the merged RunConfig per task, writes a per-task temp config, and invokes `stirrup harness --config`.
- [ ] Retained artifacts under `--output` include a `run_config.redacted.json` per task.
- [ ] `--dry-run` runs `ValidateRunConfig` on the merged config and surfaces validation errors per task.
- [ ] Existing suites (`guardrail.hcl`, `openai-responses-empty-tool-output.hcl`) continue to work unchanged.
- [ ] `docs/eval.md` is updated with the new authoring surface.
- [ ] At least one suite (e.g. update `openai-responses-empty-tool-output.hcl` to set the provider inline) demonstrates the new flow end-to-end.
- [ ] Tests cover: suite parse with file ref, suite parse with inline block, task overrides on top of suite baseline, mutually-exclusive validation, redaction in retained artifacts.

## Related

- #172 — the bug whose regression suite first exposed this gap.
- #176 — the PR that landed the suite and surfaced this issue.
- #30 — \"Turn the eval gate into a real quality signal\" — same general direction.
- `types/eval.go:51-59` — existing `RunConfigOverrides` type that this issue plumbs through.

rxbynerd:
## Summary

Closes #177.

`stirrup-eval` previously shelled out to `stirrup harness` with exactly five flags (`--prompt`, `--mode`, `--workspace`, `--trace`, `--timeout`). Provider type, model, credential federation, executor, edit strategy, verifier, code scanner, guardrail, observability, max-turns, transport — every other RunConfig field had to come from the harness binary's defaults or the operator's external environment. A suite could *describe* a regression scenario but couldn't *enforce* the configuration that made the scenario meaningful (this surfaced concretely while writing the regression suite for #172).

This PR wires the full `types.RunConfig` surface through the eval framework so suites can pin their required harness posture inline.

## What landed

Three layers of authoring surface, all optional and backwards-compatible:

1. **Suite-level external baseline** — `run_config_file = "path/to/runconfig.json"`. Loads a JSON `RunConfig` of the same shape consumed by `stirrup harness --config`. Resolved relative to the suite file's directory at load time.
2. **Suite-level inline baseline** — `run_config { ... }` HCL block, mutually exclusive with `run_config_file`. Sparse subset of `RunConfig` surfaced: `mode` (suite-level only — see Design notes), `provider { type, api_key_ref, base_url, api_key_header, query_params, region, profile, gcp_project, gcp_location }`, `model_router { type, provider, model, mode_models }`, `context_strategy { type, max_tokens }`, `edit_strategy { type, fuzzy_threshold }`, `verifier { type, command, timeout, criteria, model }`, `max_turns`.
3. **Per-task sparse overrides** — `run_config_overrides { ... }` inside a task, same shape as the inline `run_config` baseline (minus `mode`, see Design notes). Applied on top of the suite-level baseline.

The runner:

- Loads the suite baseline (file or inline) into a `types.RunConfig`.
- Applies the task's `RunConfigOverrides` on top.
- Writes the merged config to `<tmpDir>/run_config.json` (`0o600`).
- Invokes `stirrup harness --config <path> --prompt … --workspace … --trace … --timeout …`. Per the harness's documented `--config` precedence, the runner's explicit flags stay authoritative.
- When `--output` is set, persists `<output>/<suite>/<task>/run_config.redacted.json` (`0o640`) via `RunConfig.Redact()`.
- Under `--dry-run`, injects the runner's default timeout and calls `types.ValidateRunConfig` on the merged config per task, surfacing failures as failed `TaskResult`s.

A suite with no `RunConfig` and no per-task `RunConfigOverrides` falls through to today's invocation unchanged (no `--config`, no redacted artifact, no behavioural change).

## Demo

`eval/suites/openai-responses-empty-tool-output.hcl` now inlines its provider posture:

```hcl
run_config {
  provider {
    type        = "openai-responses"
    api_key_ref = "secret://OPENAI_KEY"
  }
  model_router { model = "gpt-4o-mini" }
}
```

The suite's docstring previously said *"the harness binary configured with `--provider openai-responses --model <model>`"* — that gap is exactly what #177 closes.

## Acceptance criteria mapping

All ten acceptance criteria from #177 are met (spec-compliance review found 10/10 met, evidence cited per criterion):

- [x] `eval/spec/hcl.go` parses `run_config_file`, `run_config { ... }`, `run_config_overrides { ... }`.
- [x] Mutual exclusion of `run_config_file` and `run_config` enforced at parse time.
- [x] `types.EvalSuite` and `types.EvalTask` carry the new fields; new `RunConfigSource` wrapper.
- [x] `eval/runner/runner.go` resolves the merged RunConfig per task, writes the temp config, invokes `--config`.
- [x] Retained artifacts include `run_config.redacted.json` per task.
- [x] `--dry-run` runs `ValidateRunConfig` per task and surfaces failures.
- [x] `guardrail.hcl` parses unchanged; `openai-responses-empty-tool-output.hcl` updated as the demo.
- [x] `docs/eval.md` updated with the new authoring surface.
- [x] `openai-responses-empty-tool-output.hcl` demonstrates the new flow end-to-end.
- [x] Tests cover: file ref, inline block, per-task overrides, mutual-exclusion validation, redaction in retained artifacts.

## Design notes

- **`mode` at the task-override layer is intentionally not surfaced in HCL.** The runner forwards `--mode <task.Mode>` only when the task carries an explicit top-level `mode` attribute. `RunConfigOverrides.Mode` on the Go type remains for the experiment runner (`types.Experiment.BaseConfig`). The asymmetry avoids a parse-time silent-no-op trap: a `mode` written inside an HCL overrides block would have been silently ignored at runtime under the existing harness flag-precedence rule. The doc's *Surfaced fields* table calls this out explicitly.
- **`api_key_ref` values must use the `secret://` scheme.** Validated at three layers: HCL parse (immediate, clearest error message), `ValidateRunConfig` (defense-in-depth, also catches programmatic callers), and `retainArtifacts` (stderr warning at the marshal site if either earlier gate is bypassed). Enforces the existing project invariant *"secrets never as raw values in RunConfig"* (see `CLAUDE.md`).
- **`gcp_credentials_file` deliberately not surfaced inline.** Inline file paths would need the same relative-resolution treatment as `run_config_file` itself; rather than build a second resolution path, suites needing this can use `run_config_file` pointing at a JSON config. Documented in `eval.md` and in a comment on `providerConfigSpec`.
- **`loadRunConfigFile` is duplicated from `harness/cmd/stirrup/cmd/harness.go` rather than factored.** The harness command package is binary-private. Exporting it (or inventing a new shared file-I/O helper in `types/`) would expand the public API more than the ~25-line duplication justifies. A pin-test (`TestMergeRunConfig_FileBaselineRejectsUnknownFields`) keeps the cross-module contract honest. A follow-up issue will be filed.
- **`run_config_file` is resolved relative to the suite file's directory in the spec loader.** Absolute paths pass through unchanged. The runner is pure with respect to filesystem layout — it never needs to know where the suite file lived.

## Test plan

- [x] `just build` — clean.
- [x] `just test` — all packages green. 17 new tests in `eval/runner` (38 total), 11 new tests in `eval/spec` (27 total), 2 new tests in `types`.
- [x] `just lint` — 12 diagnostics, all pre-existing on `main`. Zero new diagnostics introduced.
- [x] Existing suites (`guardrail.hcl` and the live regression suite) continue to parse and round-trip.
- [x] Per-task merge semantics covered: nil source, file-only baseline, inline-only baseline, inline + task overrides, file + task overrides, every sparse field on `RunConfigOverrides`.
- [x] Redaction guarantee covered: artifact contains `secret://[REDACTED]`, NOT the original `secret://REAL_KEY` reference; harness subprocess receives the un-redacted form.
- [x] Dry-run validation covered: validates merged config per task, surfaces failures as `fail`/`error` outcomes, preserves vacuous-pass when no RunConfig surface is set, injects default timeout so inline configs don't false-fail.
- [x] Backwards compat: a suite with no RunConfig surface produces no `--config` flag, no redacted artifact, identical behaviour to before this PR.

## Review cycle

Five reviewers ran in parallel (code, security, spec-compliance, test-coverage, api-design) and the synthesizer consolidated into a 10-blocker + 4-important remediation brief. All blockers and important findings are resolved in this PR:

| Finding | Fix |
|---|---|
| B-1 `--mode` precedence bug | Runner only forwards `--mode` when the task carries an explicit top-level `mode`; HCL grammar drops `mode` from `run_config_overrides`. |
| B-2 raw `api_key_ref` accepted | Three-layer validation (parse, validate, warn-at-marshal) requiring `secret://` scheme. |
| B-3 dry-run false-negatives on inline configs | Inject runner's default timeout before `ValidateRunConfig` (the runner's `--timeout` flag would have overridden it at runtime). |
| B-4 artifact world-readable | `run_config.redacted.json` mode tightened from `0o644` to `0o640`. |
| B-5/6/7 dry-run branch coverage | New tests for the validate-pass, merge-error, and mixed-task-config dry-run paths. |
| B-8 sub-block round-trip coverage | New test covers every nested HCL field on `runConfigOverridesSpec`. |
| B-9/10 weak test assertions | Strengthened to assert exact outcomes and to verify the harness received un-redacted input. |
| I-1 `GCPCredentialsFile` ergonomics | Documented as file-only in `eval.md` and as a code comment on `providerConfigSpec`. |
| I-2 file-baseline + task-overrides | New `TestMergeRunConfig_FileBaselineWithTaskOverrides`. |
| I-3 `loadRunConfigFile` guards | New tests pin the is-directory, size-cap, and empty-file branches. |
| I-4 `applyOverrides` nil-guard | Documented with a comment at the guard explaining the defensive contract. |

Ten lower-priority findings (mostly pre-existing concerns the review surfaced — `loadRunConfigFile` duplication consolidation, `secret://file:///` scheme allowlist, stale doc concurrency statements, `hcl.Diagnostic` error ergonomics, etc.) are intentionally deferred to follow-up issues.

### DIFF
--- a/docs/eval.md
+++ b/docs/eval.md
@@ -209,12 +209,20 @@ accepts the full `types.RunConfig` shape:
 
 | Block               | Fields                                                                                                  |
 |---------------------|---------------------------------------------------------------------------------------------------------|
-| (top-level)         | `mode`, `max_turns`                                                                                     |
-| `provider`          | `type`, `api_key_ref`, `region`, `profile`, `base_url`, `api_key_header`, `query_params`, `gcp_project`, `gcp_location` |
+| (top-level)         | `max_turns`                                                                                             |
+| `provider`          | `type`, `api_key_ref`, `region`, `profile`, `base_url`, `api_key_header`, `query_params`, `gcp_project`, `gcp_location` (`gcp_credentials_file` is intentionally absent — file paths inline need file-relative resolution semantics; use `run_config_file` with a full JSON config when this field is needed) |
 | `model_router`      | `type`, `provider`, `model`, `mode_models`                                                              |
 | `context_strategy`  | `type`, `max_tokens`                                                                                    |
 | `edit_strategy`     | `type`, `fuzzy_threshold`                                                                               |
 | `verifier`          | `type`, `command`, `timeout`, `criteria`, `model`                                                       |
+
+Mode is controlled by the task-level `mode = "..."` attribute (or the
+suite-level `mode` baked into a `run_config_file`), not by a
+`mode = "..."` setting inside a `run_config { ... }` or
+`run_config_overrides { ... }` block. The runner forwards the task's
+`mode` to the harness via the `--mode` flag, which always wins over
+the merged `--config` payload, so surfacing `mode` inside the inline
+block would be an authoring trap.
 
 **Redaction.** When `--output` is set, the runner writes the redacted
 form of the merged config to `<output>/<suite>/<task>/run_config.redacted.json`
--- a/docs/eval.md
+++ b/docs/eval.md
@@ -106,6 +106,133 @@ generate suites from a higher-level tool and emit the static HCL.
 Suite definitions live in `eval/suites/`. CI baselines live in
 `eval/baselines/`.
 
+### Pinning the harness configuration per suite
+
+Some regression suites are only meaningful against a specific harness
+configuration — a particular provider adapter, a specific credential
+federation mode, an executor with a particular network posture, a
+non-default verifier. Encoding that posture in the suite itself,
+rather than in operator memory or a wrapper script, keeps the suite
+self-describing and prevents silent skew between "what the author
+intended" and "what the runner actually passed on the command line".
+This is the surface introduced by issue #177.
+
+A suite can declare a `RunConfig` baseline in one of two
+mutually-exclusive forms, and each task can carry sparse overrides on
+top:
+
+```hcl
+suite "fix-nil-check-regressions" {
+  description = "..."
+
+  # Form A — external file. Path is resolved relative to the suite
+  # file's directory at load time, so authors can use intuitive
+  # relative paths (e.g. "configs/base.json" next to the suite).
+  run_config_file = "configs/openai-responses-base.json"
+
+  # Form B — inline. Mutually exclusive with run_config_file.
+  # run_config {
+  #   mode      = "execution"
+  #   max_turns = 10
+  #   provider {
+  #     type        = "openai-responses"
+  #     api_key_ref = "secret://OPENAI_KEY"
+  #   }
+  #   model_router {
+  #     type  = "static"
+  #     model = "gpt-4o-mini"
+  #   }
+  # }
+
+  task "task-001" {
+    description = "..."
+    mode        = "execution"
+    prompt      = "..."
+
+    # Per-task overrides apply on top of the suite baseline. Only set
+    # the fields that should differ for this task; nil/empty fields
+    # leave the baseline untouched.
+    run_config_overrides {
+      max_turns = 4
+      provider {
+        type        = "anthropic"
+        api_key_ref = "secret://ANTHROPIC_API_KEY"
+      }
+    }
+
+    judge { type = "file-exists" paths = ["done.txt"] }
+  }
+}
+```
+
+The carrier types are `EvalSuite.RunConfig` (`*RunConfigSource`) and
+`EvalTask.RunConfigOverrides` (`*RunConfigOverrides`) in
+[`types/eval.go`](../types/eval.go). The HCL grammar for both blocks
+lives in [`eval/spec/hcl.go`](../eval/spec/hcl.go); the runner merge
+and harness invocation live in
+[`eval/runner/runner.go`](../eval/runner/runner.go).
+
+**Precedence.** The runner merges in three layers, lowest to highest:
+
+1. The suite baseline (from `run_config_file` if set, else from the
+   inline `run_config { ... }` block).
+2. The per-task `run_config_overrides` block, applied as a sparse
+   patch on top of the baseline.
+3. The runner-supplied harness flags — `--prompt`, `--mode`,
+   `--workspace`, `--trace`, `--timeout` — passed explicitly when the
+   runner shells out to `stirrup harness`. These win over any value
+   carried by the merged RunConfig, matching the harness's documented
+   `--config` precedence (see the `--config` flag docstring in
+   [`harness/cmd/stirrup/cmd/harness.go`](../harness/cmd/stirrup/cmd/harness.go)).
+
+The merged config is written to `<task tmpDir>/run_config.json` and
+passed as `--config <path>` to the harness. Suites with neither
+`run_config_file` nor an inline `run_config` block, and tasks with no
+`run_config_overrides`, fall through to the legacy invocation — no
+`--config` flag, no merged artifact.
+
+**Path resolution.** `run_config_file` is resolved relative to the
+suite file's directory at load time; absolute paths are preserved
+verbatim. This keeps the runner pure: it never needs to know where the
+suite file was loaded from.
+
+**Mutual exclusion.** A suite that sets both `run_config_file` and an
+inline `run_config { ... }` block is rejected at load time. The merge
+order would otherwise be ambiguous.
+
+**Surfaced fields.** The HCL grammar currently surfaces the fields
+listed below. Operators who need a less-common field (Gemini safety
+settings, full credential federation blocks, dynamic-router cheap/
+expensive overrides, composite verifiers) should keep the baseline in
+a JSON file and reference it via `run_config_file` — the JSON loader
+accepts the full `types.RunConfig` shape:
+
+| Block               | Fields                                                                                                  |
+|---------------------|---------------------------------------------------------------------------------------------------------|
+| (top-level)         | `mode`, `max_turns`                                                                                     |
+| `provider`          | `type`, `api_key_ref`, `region`, `profile`, `base_url`, `api_key_header`, `query_params`, `gcp_project`, `gcp_location` |
+| `model_router`      | `type`, `provider`, `model`, `mode_models`                                                              |
+| `context_strategy`  | `type`, `max_tokens`                                                                                    |
+| `edit_strategy`     | `type`, `fuzzy_threshold`                                                                               |
+| `verifier`          | `type`, `command`, `timeout`, `criteria`, `model`                                                       |
+
+**Redaction.** When `--output` is set, the runner writes the redacted
+form of the merged config to `<output>/<suite>/<task>/run_config.redacted.json`
+alongside the trace. `secret://...` references are rewritten to
+`secret://[REDACTED]` by `RunConfig.Redact()`. This matches the
+project-wide invariant in [`CLAUDE.md`](../CLAUDE.md) that secrets
+never appear in `RunConfig` artifacts persisted alongside traces or
+recordings. Operators may safely audit, attach, and commit the
+redacted artifact.
+
+**Dry-run.** `eval run --dry-run` runs `types.ValidateRunConfig` on
+the merged config for every task. A task whose merged config fails
+validation surfaces as a failed `TaskResult` with the validation error
+in `Reason`; tasks whose merged config validates produce the synthetic
+all-pass `TaskResult` the dry-run path always emits. Suites without
+any RunConfig surface skip the validation step entirely and produce
+the same synthetic pass result as before.
+
 ### Judges
 
 A judge decides whether a task passed by inspecting the workspace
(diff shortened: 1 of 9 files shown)


rxbynerd:
Absorbed into #186

PR #191, opened 2026-05-15
provider/gemini: disable streamFunctionCallArguments for 3.x compatibility

rxbynerd:
## Summary

A live Vertex AI run against the new `gemini-3.1-pro-preview` model fails immediately with:

```
{"level":"ERROR","msg":"provider stream failed","provider":"gemini","model":"gemini-3.1-pro-preview","error":"functionCall part missing name"}
```

### Reproduction

```
GOOGLE_APPLICATION_CREDENTIALS="<service-account-key.json>" \
  ./stirrup harness \
    --prompt "Give me a summary of the safety rings doc" \
    --provider gemini \
    --gcp-project rubynerd-net \
    --model gemini-3.1-pro-preview \
    --workspace .
```

### Root cause

Under `toolConfig.functionCallingConfig.streamFunctionCallArguments=true`, the Gemini 3.x family streams a function call across multiple SSE chunks where only the first carries `name`. Continuation chunks deliver `partialArgs` as an **array of JSON-path delta records** (`{jsonPath, stringValue, willContinue}`) rather than the cumulative JSON-object snapshots the 2.x family emitted. A representative first continuation chunk from a captured Vertex response:

```
data: {"candidates":[{"content":{"role":"model","parts":[{"functionCall":{"partialArgs":[{"jsonPath":"$.path","stringValue":"docs/safety-rings.md","willContinue":true}],"willContinue":true}}]}}], ...}
```

The existing parser in `harness/internal/provider/gemini.go` keys per-call buffers by `fc.Name` and reads `partialArgs` as a cumulative JSON object. The first continuation chunk has an empty `Name` and trips the explicit `"functionCall part missing name"` guard.

### Why not implement the new delta format

Disabling the streamed-args flag is the right fix because:

- With the flag off, both `gemini-2.5-pro` and `gemini-3.1-pro-preview` emit a single `functionCall` part with `name` and `args` populated in one SSE chunk (verified empirically against `rubynerd-net`). The existing parser already handles that shape uniformly.
- Supporting the new delta format would require index-keyed slot correlation plus JSON-path synthesis on the receive side, for no operational upside — the harness emits its trace post-turn, not mid-stream, so per-chunk argument visibility is not load-bearing.
- The parser-side tolerance for `partialArgs` and the `"missing name"` guard are retained so the adapter remains correct if a future operator re-enables the flag.

### Changes

- `harness/internal/provider/gemini_request.go` — `StreamFunctionCallArguments: true` → `false`.
- `harness/internal/provider/gemini_types.go` — rewrote the `geminiToolConfig` rationale comment and updated the `geminiFunctionCall.PartialArgs/WillContinue` note to reflect that those fields are no longer exercised on the happy path.
- `harness/internal/provider/gemini_request_test.go` — renamed `TestBuildGenerateContentRequest_StreamFunctionCallArgumentsTrueWhenToolsPresent` → `...FalseWhenToolsPresent` and inverted the assertion.
- `harness/internal/provider/gemini_test.go` — added `TestGeminiAdapter_NonStreamedFunctionCall3xShape`, a regression test that replays the Gemini 3.x non-streamed shape (single SSE chunk with `name + args` populated alongside `finishReason: STOP`) and pins `STOP -> tool_use` promotion.

## Validation

- [x] `just test` — all packages pass.
- [x] `golangci-lint run ./harness/internal/provider/...` — 0 issues.
- [ ] Manual smoke against Vertex AI with the reproduction command above (maintainer, requires GCP creds).

## Follow-up

Not in this PR: Gemini 3.x parts carry `thoughtSignature` blobs that operators should round-trip on subsequent turns to preserve the model's hidden chain-of-thought across tool exchanges. The harness currently drops them on receive and never re-sends them. Worth tracking as a separate change.

### DIFF
--- a/harness/internal/provider/gemini.go
+++ b/harness/internal/provider/gemini.go
@@ -288,7 +288,9 @@ type toolCallBuf struct {
 //   - FunctionCall parts arrive in two patterns:
 //     1) Streamed: one or more chunks with WillContinue=true and
 //     PartialArgs, then a final chunk with WillContinue=false and either
-//     Args or PartialArgs populated.
+//     Args or PartialArgs populated. Pattern 1 is handled defensively;
+//     with streamFunctionCallArguments=false the production path is
+//     always pattern 2.
 //     2) Single: one chunk with neither WillContinue nor partial markers,
 //     just Args.
 //   - finishReason on Candidates[0] terminates the turn — emit
@@ -463,6 +465,7 @@ func (g *GeminiAdapter) consumeSSE(
 							buf.args = snapshot
 						}
 
+						// Unreachable when streamFunctionCallArguments=false (current default); retained for correctness if the flag is re-enabled.
 						if !fc.WillContinue {
 							// Final chunk for this call: emit, then
 							// advance the sequence counter used for ID
--- a/harness/internal/provider/gemini_request.go
+++ b/harness/internal/provider/gemini_request.go
@@ -87,7 +87,7 @@ func BuildGenerateContentRequest(
 		req.ToolConfig = &geminiToolConfig{
 			FunctionCallingConfig: geminiFunctionCallingConfig{
 				Mode:                        "AUTO",
-				StreamFunctionCallArguments: true,
+				StreamFunctionCallArguments: false,
 			},
 		}
 	}
--- a/harness/internal/provider/gemini_request_test.go
+++ b/harness/internal/provider/gemini_request_test.go
@@ -326,7 +326,14 @@ func TestBuildGenerateContentRequest_CustomSafetySettings(t *testing.T) {
 	}
 }
 
-func TestBuildGenerateContentRequest_StreamFunctionCallArgumentsTrueWhenToolsPresent(t *testing.T) {
+// TestBuildGenerateContentRequest_StreamFunctionCallArgumentsFalseWhenToolsPresent
+// pins the request shape: when tools are declared the adapter sets
+// functionCallingConfig.mode="AUTO" and streamFunctionCallArguments=false.
+// The flag stays off because Gemini 3.x's streamed-args wire format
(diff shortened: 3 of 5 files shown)


rxbynerd:
First-round review remediation pushed as fa7ad9d.

**Applied this round:**
- **M1** — `omitempty` on `geminiFunctionCallingConfig.StreamFunctionCallArguments` (`gemini_types.go`). Omission is now wire-equivalent to explicit `false`, matching the rest of the struct and Vertex's server default. Existing assertion in `TestBuildGenerateContentRequest_StreamFunctionCallArgumentsFalseWhenToolsPresent` continues to pass — the test decodes into a struct (not a substring search), so a zeroed `bool` from an omitted field satisfies the `!stream` check.
- **M2 (location 1)** — `consumeSSE` doc comment now notes pattern 1 is defensive: with `streamFunctionCallArguments=false` the production path is always pattern 2.
- **M2 (location 2)** — added a one-line annotation above `if !fc.WillContinue` flagging the surrounding continuation handling as retained for correctness if the flag is ever re-enabled.

**Deferred:**
- **D1** (end-to-end pin of the stream flag) — moot once `omitempty` lands, since the field is no longer emitted on the wire at all.
- **D2** (`TestGeminiAdapter_BlankFunctionCallName` for the empty-name guard at `gemini.go:438-441`) — pre-existing coverage gap, not introduced by this PR. Filing as a follow-up issue.
- **D3 / D4** — cosmetic; no action.

**Validation:** `just test` green, `golangci-lint run ./harness/internal/provider/...` clean. Repo-wide `just lint` still surfaces the same pre-existing issues across unrelated files (`harness_test.go`, `factory.go`, `runconfig.go`, `stirrup-resolve/*`) that were present before this branch — none touched here.

Issue #192, opened 2026-05-15
provider: per-model request-shape overrides for upstream wire-format drift


rxbynerd:
## Problem

Stirrup intentionally targets a wide spectrum of LLM providers via thin, hand-rolled HTTP+SSE adapters (`harness/internal/provider/*.go`). The implicit assumption baked into each adapter is that **a provider's wire format is stable across all model versions under that provider**. That assumption is false in practice, and the divergence is starting to bite:

### Evidence point 1 — Vertex AI Gemini 3.x (resolved in #191)

Vertex AI changed the wire format for streamed function-call arguments between Gemini 2.x and 3.x:

- `streamFunctionCallArguments=true` on 2.x: each chunk carries the function `name` plus a cumulative JSON-object snapshot in `partialArgs`.
- `streamFunctionCallArguments=true` on 3.x: only the first chunk carries `name`; continuation chunks omit it and emit `partialArgs` as an *array of JSON-path delta records* (`[{jsonPath, stringValue, willContinue}]`). The final chunk is an empty `functionCall: {}`.

The harness assumed the 2.x shape and aborted with `"functionCall part missing name"` on the first 3.x continuation. We worked around it by setting `streamFunctionCallArguments=false` for *all* Gemini models, sacrificing per-chunk trace visibility. The same single `gemini` adapter now serves both 2.5 and 3.x — only by avoiding the diverging surface.

### Evidence point 2 — OpenAI-compatible `max_tokens` → `max_completion_tokens` (open)

The `openai-compatible` adapter (Chat Completions) hard-codes the legacy field name:

```go
// harness/internal/provider/openai.go:106
type openaiRequest struct {
    ...
    MaxTokens   int  \`json:\"max_tokens\"\`
    ...
}
```

OpenAI deprecated `max_tokens` in favour of `max_completion_tokens` for the o1 model family (and newer reasoning models). **Azure OpenAI surfaces this break**: deployments backed by the newer model lineages reject requests with `max_tokens` set and require `max_completion_tokens`. The same adapter, the same provider type (\`openai-compatible\`), the same `--base-url`, two mutually incompatible wire requirements split along model lines.

### Generalised shape of the problem

These two cases are not isolated. The pattern is:

> **Provider P offers models M1 and M2. The adapter for P sends one wire shape S. M1 accepts S; M2 rejects S or interprets S differently. The harness's only knob is provider-level, not (provider, model)-level.**

[...]

Issue #160, opened 2026-05-10
smoke tests: OpenAI via Azure OpenAI


rxbynerd:
## Background

#130 / #132 shipped `.github/workflows/smoke-anthropic.yml`, a manually-dispatched workflow that exercises the Anthropic Workload Identity Federation (WIF) credential path end-to-end against the live Anthropic API with **no static API key in scope**. It uses GitHub's OIDC token, exchanges it via `--anthropic-from-github-actions`, and verifies the run via two independent assertions (trace outcome + workspace-file contents).

#118 shipped `azure-workload-identity` as the credential source for Azure OpenAI / Foundry: GHA OIDC JWT → Microsoft Entra `client_credentials` + JWT-bearer assertion → short-lived bearer token on `Authorization: Bearer …` against the Azure OpenAI deployment. The shape of an `examples/runconfig/azure-openai-wif-github-actions.json` fixture and the operator walkthrough at [`docs/azure-workload-identity.md`](../blob/main/docs/azure-workload-identity.md) (Section 2) are already in tree.

This issue lands the **smoke test** equivalent for Azure OpenAI — a manually-dispatched CI workflow that proves the federation path works against a live Azure OpenAI deployment. Bedrock and Vertex are sister issues #161 and #162; setup work shared across all three (e.g. naming conventions, verify-step pattern) should land here first since this issue is the first cross-cloud entrant after Anthropic.

## Goal

A `workflow_dispatch`-only `.github/workflows/smoke-azure-openai.yml` workflow that:

1. Dials Azure OpenAI via Entra ID Workload Identity Federation (no static client secret, no API key).
2. Runs the harness for ≤ 5 turns against a cheap Azure OpenAI deployment (e.g. `gpt-5.4-nano`, which is already provisioned on the stirrup test resource — see "Provisioned state" below).
3. Verifies the run succeeded (trace outcome + workspace artifact).
4. Uploads the trace as a debugging artifact (7-day retention).
5. Has a companion operator-facing setup guide that an Azure tenant admin can follow end-to-end with no prior context.

## Workflow shape

`.github/workflows/smoke-azure-openai.yml`. Mirrors the structure of `smoke-anthropic.yml`:

```yaml
name: Smoke Test - Azure OpenAI WIF

# End-to-end smoke test for Azure OpenAI Entra ID Workload Identity
# Federation. Manually dispatched. Exchanges the GitHub Actions OIDC
# token for a Microsoft Entra access token via the federated identity
[...]

PR #196, opened 2026-05-15
docs: design plan for per-model request-shape overrides (#192)

rxbynerd:
## Summary

Adds `docs/provider-quirks.md`, the architectural contract for issue
#192 — per-model wire-format drift inside a single provider type
(e.g. OpenAI `max_tokens` vs `max_completion_tokens` across the
o-series, Gemini 3.x `streamFunctionCallArguments` semantics). The
plan establishes the registry shape, where adapters consult it,
forward-compat defaults, v1 scope, and the wave-by-wave rollout
subsequent PRs will deliver against.

This PR is doc-only. It does not change any runtime behaviour, nor
does it touch any adapter; the implementation lands in follow-up
PRs sequenced against the rollout plan in section 9.

## Key design decisions

- **`ProviderQuirks` + `Registry`**: a typed struct (field
  renames, omissions, value overrides, per-adapter structural
  flags) resolved per-stream against an ordered rule registry keyed
  on `(provider-type, model-glob)`.
- **Specificity-then-declaration-order composition**: a wide rule
  (`gpt-5*`) plus a narrower override (`gpt-5-chat*`) compose
  cleanly without depending on authors to maintain a brittle
  declaration order.
- **The Codec invariant**: one resolved `ProviderQuirks` drives
  both the send and parse paths. Lifted from the prior-art survey
  — LiteLLM, LangChain, and Vercel AI SDK all ship this invariant.
- **Forward-compatible defaults**: new model ids inherit the
  latest known wire shape; older families opt in to legacy shape
  via explicit rules.
- **Read-only operator surface in v1**: a `stirrup providers
  quirks` introspection subcommand and a `provider.quirk.applied`
  span attribute. Operator-authored overrides are deliberately
  deferred.
- **Bedrock excluded from v1**: the SDK already absorbs the
  model-family divergences known today; the registry remains
  provider-agnostic so Bedrock can join later without redesign.

## Process

- Prior-art survey (LiteLLM, LangChain `langchain_openai`, Vercel
  AI SDK, OpenRouter, vLLM) captured under `reviews/issue-192-prior-art.md`
  and cited in section 1.1 of the plan.
- One `technical-doc-reviewer` pass; all four flagged Critical
  issues remediated:
  - rule-composition contradiction between the `Rule` doc-comment
    and the composition subsection (lifted to specificity-then-
    declaration-order);
  - `Resolve` map-initialisation contract pinned;
  - `mapBedrockStopReason` cross-reference corrected;
  - section 3.3 cached-adapter-quirks claim aligned with section
    4 per-stream resolution.
  Voice violations (4) and most Important issues remediated;
  remaining items called out as open questions in section 10.

## Test plan

- [ ] Read sections 1–4 and 9 to validate the architectural choice.
- [ ] Confirm the rollout sequencing in section 9 lines up with
      release planning (Wave 2 unblocks issue #160's smoke test).
- [ ] Confirm the open questions in section 10 (especially
      Bedrock model-id normalisation, eval-suite implications,
      and the `runs replay` slice-attribute rendering) — anything
      that needs answering before Wave 1 lands.

Refs #192.

### DIFF
--- a/docs/provider-quirks.md
+++ b/docs/provider-quirks.md
@@ -1,0 +1,945 @@
+# Provider quirks: per-model request-shape overrides
+
+**Status:** Design plan. Tracks [issue #192].
+
+[issue #192]: https://github.com/rxbynerd/stirrup/issues/192
+
+This document specifies the framework Stirrup will adopt to handle
+divergences between models served under the same provider type. It
+covers the registry shape, where it integrates with the existing
+adapters, how it stays visible to operators, what ships in v1, and
+which divergences are deferred. Implementation is out of scope here;
+this is the architectural contract that subsequent waves will
+implement against.
+
+## 1. Problem
+
+Stirrup ships five provider adapters
+([`harness/internal/provider/*.go`](../harness/internal/provider/)).
+Each adapter assumes a single wire shape applies to every model under
+that provider. Recent breakage shows the assumption is false:
+
+| Provider | Model pattern | Divergence | Status |
+|---|---|---|---|
+| `gemini` (Vertex AI) | `gemini-3*` | `streamFunctionCallArguments` deltas; `thoughtSignature` on parts | Worked around in [#191] (flag off); thoughtSignature deferred |
+| `openai-compatible` | OpenAI o-series; Azure OpenAI deployments backed by newer reasoning models | `max_tokens` rejected, `max_completion_tokens` required | **Broken** |
+| `openai-compatible` | OpenAI reasoning models | `temperature` rejected on some o-series deployments | Suspected |
+| `openai-responses` | Various Azure Foundry deployments | Required-key rules for empty `output` / `text` fields | Resolved per-case in [#172] / [#176] |
+
+[#191]: https://github.com/rxbynerd/stirrup/pull/191
+[#172]: https://github.com/rxbynerd/stirrup/issues/172
+[#176]: https://github.com/rxbynerd/stirrup/issues/176
+
+The generalised shape: provider P serves models M1 and M2 under one
+adapter wire shape S. M1 accepts S; M2 rejects S or interprets S
+differently. The harness's only existing knobs (`--query-param`,
+`--api-key-header`, `--base-url`) live at the URL/header layer; they
+do not address body shape, field renames, conditional field omission,
+or value reshaping.
+
+Two non-starters frame the design space:
+
+- **Per-model adapters.** N providers × M models is an explosion the
+  hand-rolled stdlib-HTTP philosophy cannot absorb. Five adapters of
+  several hundred lines is the budget.
+- **Wire-shape knobs on `RunConfig`.** `RunConfig` is a description of
+  *intent* (provider, model, max tokens), not a request-shape patch.
+  Bleeding wire fields onto the operator surface would mean operators
+  routinely encode adapter internals, which both defeats Stirrup's
+  declarative-config philosophy and creates a permanent backwards-
+  compatibility burden as quirks shift.
+
+## 1.1 Prior art
+
+A focused survey of comparable projects (LiteLLM, LangChain
+`langchain_openai`, Vercel AI SDK, OpenRouter, vLLM) shows the
+industry has converged on a **capability table plus small transform
+hooks** pattern. Where projects differ is in keying strategy and
+parse-path organisation. Full research packet:
+[`reviews/issue-192-prior-art.md`](../reviews/issue-192-prior-art.md).
+
+Headline findings that shape this plan:
+
+- **LiteLLM** uses a class-per-family hierarchy (`OpenAIOSeriesConfig`
+  extends `OpenAIGPTConfig`) with a hand-edited dispatch ladder.
+  Power: high; readability: low; new-model lag in the issue tracker
+  is a recurring complaint.
+- **LangChain** splits data and behaviour: a `_PROFILES` dict keyed
+  by exact model id, a sibling TOML overrides file, and a separate
+  `_compat.py` module for response-shape converters.
+- **Vercel AI SDK** is the cleanest expression: a single typed
+  capability struct, a `getOpenAILanguageModelCapabilities(modelId)`
+  function using `startsWith` prefix chains, and inline conditionals
+  in the adapter. Defaults to the modern shape; new models are
+  forward-compatible by default.
+- **vLLM** is illustrative of the server-side mirror: it accepts
+  both `max_tokens` and `max_completion_tokens` via Pydantic
+  deprecation and prefers the new one — "tolerant inbound, strict
+  outbound." Azure OpenAI's strict outbound is what the harness
+  absorbs.
+- **OpenRouter** is a negative result: no public wire-format
+  rewrite spec; everything is implicit per upstream.
+
+Two invariants from the survey are load-bearing for the design that
+follows:
+
+- **One source of truth for send and parse.** LiteLLM's `*Config`
+  class hosts both `transform_request` and `chunk_parser`; LangChain
+  ties `_compat.py` converters to the same profile entry. The cost of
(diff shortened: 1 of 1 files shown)


rxbynerd:
PR #198 (issue #197 — openai-compatible retry policy) merged into `gh-197` and is up for review in parallel with this plan. The two changes are orthogonal — retry handling sits at the HTTP-execution layer (`DoWithRetry`), quirks sit at the request-marshalling layer — but there are a handful of small coordination items that should be folded into this doc before Wave 2 lands.

I do **not** think any of these require a substantive design change; they're either wording precision or sequencing notes for the implementer.

### 1. Section 3.4 "Operator surface" — tighten the "RunConfig is unchanged" claim

The doc says (line 281-ish):

> `RunConfig` is unchanged. Operators do not author quirks.

After #198, `RunConfig.ProviderConfig` carries an operator-authored `Retry *ProviderRetryConfig` block (retry policy, hard ceilings, defaults applied in `ValidateRunConfig`). That is execution behaviour, not request shape, so the *spirit* of the claim still holds — but the literal phrasing is no longer accurate. Suggest:

> `RunConfig`'s **wire-shape surface** is unchanged. Operators do not author quirks. (Retry policy is configurable per provider, but is an execution-layer concern, not a wire-shape patch.)

### 2. Section 4 / Wave 2 (section 9) — `Stream` call ordering

Post-#198 the `openai-compatible` `Stream` body looks like:

```
build openaiRequest{} → json.Marshal → resolveBearer → composeURL → http.NewRequestWithContext → DoWithRetry → consumeSSE
```

(see `harness/internal/provider/openai.go:328-378`)

Wave 2's `registry.Resolve(...)` insertion belongs **before** the marshal step (so the resolved quirks can drive `FieldRenames` / `OmitFields` / `ValueOverrides`), and the resulting body bytes feed into `http.NewRequestWithContext` exactly as today. `DoWithRetry` reads the body once and replays it from `req.GetBody` (which `strings.NewReader` sets automatically) — so quirks resolution is correctly **once per `Stream` call**, not once per retry attempt. Worth a one-line note in section 4 acknowledging this: "with retry, the body is constructed once from the resolved quirks; `DoWithRetry` replays the same bytes verbatim across attempts."

### 3. Section 3.3 / 3.4 — adapter struct surface has grown

`OpenAICompatibleAdapter` now carries:
- `RetryPolicy` (constructor arg, value)
- `Logger *slog.Logger` (post-construction field; nil-safe; factory wires the `ScrubHandler`-backed logger)
- `Tracer` and `Metrics` (existing, post-construction fields)

When Wave 2 introduces DI for the quirks registry (section 3.2 mentions "tests may swap in a different registry via dependency injection"), I'd suggest following the **post-construction field** pattern (capitalised public field, nil = use package default `BuiltinRules()`) for consistency with `Logger`/`Tracer`/`Metrics` rather than extending the positional constructor signature again. The constructor at `openai.go:77` is already a four-argument signature post-#198.

### 4. Wave 5 (section 9) — `docs/configuration.md` already has a Retry policy section

#198 added a `### Retry policy` subsection to `docs/configuration.md` (config-file keys + the four `--provider-retry-*` flags + the default table). When Wave 5 lands its "per-model wire-shape decisions are registry-driven and not operator-configurable" note, place it as a sibling subsection rather than rewriting the area. Suggested adjacent header: `### Per-model wire-shape (quirks)`.

### 5. Section 7.1 — fixture comparison is unaffected by the retry helper

The captured `request.json` fixtures compare against the bytes the marshaller produces. `DoWithRetry` replays the same byte slice on every attempt (no per-attempt re-marshal). Worth a parenthetical reassurance in section 7.1 — implementer reviewing this alongside #198 will want confirmation that the retry seam doesn't perturb the fixture contract.

### 6. Observability coexistence (section 8)

Both PRs attach to the `provider.stream` span. The combined operator-facing trace for an `o1-mini` run that hit a transient 429 will carry:

- `provider.quirk.applied = ["OpenAI o1 family requires max_completion_tokens"]` (from quirks)
- `http.status_code` (existing)
- `provider_retry_attempt` event (intermediate retry, from #198)
- `rate_limited` event only on terminal 429 (preserved by #198)
- Counter `stirrup.harness.provider_retry_outcomes` (from #198)

No namespace collision; worth a one-liner in section 8 noting that quirks attributes live on the same span as the existing retry/HTTP attributes so operators see a single coherent picture.

---

**TL;DR:** No design conflict. Section 3.4's "RunConfig is unchanged" wording needs a small clarification (it's the wire-shape surface that's unchanged, not the whole struct), Wave 2's implementer guidance should call out the new ordering at the top of `Stream`, and Wave 5's docs pass needs to land next to the existing Retry policy section rather than over it.

rxbynerd:
**Wave 2 design approved — clobbering markers**

The `api-design-advisor` pass over this PR and issue #221 has produced an approved design at `stirrup-coord/scratch/wave-2-design.md` (local scratch, not committed). Implementation of Wave 2 proceeds against that doc.

**Sections kept:** §1, §1.1, §2, §3.1 (shape retained; `StructuralFlags any` replaced by typed `BehaviourFlags`), §3.2–3.5, §4, §5, §6, §7, §8, §10, §11.

**Sections amended:**
- `StructuralFlags any` → `BehaviourFlags ProviderBehaviourFlags` (typed, per-adapter nested struct; removes runtime `any` assertions)
- Reasoning-class OmitFields (7 entries) → `OmitSamplingParams bool` on `OpenAIBehaviourFlags`
- `CompatProfile string` added to `ProviderConfig` (narrow enum; only legal value in v1: `"zai-glm"`) — this is the Z.ai injection mechanism and is not a violation of the RunConfig-as-intent invariant
- Types live in a sub-package `harness/internal/provider/quirks/` rather than a file in the provider package (cleaner import boundary for tests and compat packages)
- Canonical-field constants stay per-adapter (not consolidated in `quirks/fields.go`), avoiding an import cycle

**Sections replaced:**
- §9 rollout plan — superseded by wave-2-design.md §7 (the current main already has `max_completion_tokens` hard-coded and `Temperature *float64 omitempty`, so the "Wave 2 = OpenAI integration" framing shifts to reasoning-param omissions + Z.ai compat + scaffolding)
- §12 implementation summary file list — superseded by wave-2-design.md §7

**Key change from prior design for Z.ai:** Z.ai/GLM lives at `harness/internal/provider/compat/zai/` and exposes `CompatRule() quirks.Rule`. It is not in `BuiltinRules()`; it is injected when `ProviderConfig.CompatProfile = "zai-glm"`.

The `docs/provider-quirks.md` added by this PR will be updated to the approved design in the Step 6 documentation pass. The prior-art research in §1.1 is retained and cited.

rxbynerd:
Superseded by the landed implementation. The capability/quirks layer this design proposed (reconciled with #221's "capability profile" framing during the Wave 2 api-design pass) shipped across Waves 2–5 as `harness/internal/provider/quirks/` — `ProviderQuirks` + `Registry` keyed on `(provider-type, model-glob)`, with the Codec-invariant intent realised for the OpenAI Chat, Anthropic, and Gemini adapters. The in-tree design record is now `docs/provider-quirks.md`.

The prior-art survey and the `ProviderQuirks`/`Registry` shape here were load-bearing inputs; what diverged from this doc: capabilities became top-level fields on `ProviderQuirks` (`ToolChoice`, `StructuredToolResults`, and the #222 `ParallelToolCalls`/`ToolExamples`) rather than living only under per-provider behaviour flags, and Z.ai landed as a `compat/zai` rule injection rather than a first-party profile.

Umbrella #192 is now closed (the capability layer exists). Two residuals are tracked separately: #332 (model the OpenAI Responses wire divergences as quirks rules — the Codec invariant does not yet hold for Responses) and #333 (re-enable Gemini 3.x per-chunk streaming, gated on a live capture). Recommend closing this PR in favour of the in-tree doc; leaving that call to you.

PR #251, opened 2026-05-21
cli: add run-config subcommand and --output-runconfig

### LINKED ISSUE #240: RunConfig generator (& accept RunConfig from stdin)
## Background

stirrup is a coding-agent harness designed for autonomous operation, but
during development and prototyping it is regularly touched by a human
building or testing a workflow. Today it offers two ways to provide a
`RunConfig`:

- A JSON file via `--config <path>` (mirrors
  [`proto/harness/v1/harness.proto`](proto/harness/v1/harness.proto) and
  [`types/runconfig.go`](types/runconfig.go))
- Individual CLI flags on `stirrup harness` (see
  [`docs/configuration.md`](docs/configuration.md))

The two surfaces compose: a `--config` file is the base, explicit flags
override individual fields, defaults do not override the file
(`harness/cmd/stirrup/cmd/harness.go:807` — `applyOverrides`). What is
missing is a way to *capture* the result, or to compose configs
incrementally without writing a JSON file at every step.

This proposal adds a new `stirrup run-config` subcommand whose only job
is to produce a fully-resolved `RunConfig` document, and extends
`stirrup harness` to accept that document on stdin. Together they
enable a UNIX-style pipeline where each stage layers one more
adjustment on top of the previous one before the final stage runs the
agent:

```sh
stirrup run-config --model claude-opus-4-7 \
  | stirrup run-config --max-turns 100 \
  | stirrup run-config --mode execution --executor container \
  | stirrup harness --prompt "refactor X"
```

This is strictly additive: the existing `--config` flag, the existing
flag-only path, and the existing precedence rules are unchanged.

## Goals

1. **`stirrup run-config` subcommand.** Accepts every CLI flag that
   produces a `RunConfig` field, plus a base config on stdin or via
   `--config`, and emits the resolved `RunConfig` as JSON to stdout.
   Does not run the agentic loop.
2. **`stirrup harness` reads stdin.** When stdin carries a JSON
   `RunConfig` (either via `--config -` or via auto-detection on a
   non-TTY stdin) it is treated identically to a `--config <path>`
   base, with explicit flags layered on top using the existing
   precedence chain.
3. **`stirrup harness --output-runconfig <path>`.** After the
   resolved `RunConfig` is built and validated, write it as JSON to
   `<path>` (use `-` for stdout) and exit without running. Equivalent
   to a dry-run that captures the exact config that *would* have run.
4. **Shared builder.** The three call sites (`run-config`, `harness`
   with `--config`, `harness` with flags only) converge on a single
   builder helper so flag overrides, prompt resolution, and source
   precedence cannot drift.

## Non-goals

- No new `RunConfig` schema fields. This is a CLI/UX feature only.
- No change to the gRPC `TaskAssignment` payload or `stirrup job`
  entrypoint — the control plane already speaks JSON `RunConfig`.
- No interactive editor / TUI for editing configs.
- No "delta" output that shows only operator-set fields versus
  defaults — `run-config` always emits the merged, resolved document.
- No removal of any existing flag.

### New: `stirrup run-config`

```text
stirrup run-config [flags]

Produce a fully-resolved RunConfig JSON document and print it to stdout.
Does NOT run the agentic loop. Composable via UNIX pipes: stdin is
treated as a base RunConfig, then explicit flags override individual
fields, mirroring how `stirrup harness --config <file>` already works.

Resolution order (lowest → highest precedence):
  1. Defaults (flag DefValues, mode-derived defaults)
  2. Base RunConfig from stdin OR --config <path> (mutually exclusive)
  3. Explicit flags (those whose Changed() bit is set)

Flags: identical to `stirrup harness` for every RunConfig-producing
flag (see `--help`). CLI-only behaviour flags (--export-workspace-
required) are excluded because they do not map to RunConfig fields.

Output:
  - Default: pretty-printed JSON (2-space indent) to stdout
  - --compact: single-line JSON
  - --redact: apply RunConfig.Redact() before emit (rewrites
    secret:// references to secret://[REDACTED] — produces a copy
    safe to commit / share, but not runnable as-is)

Exit codes:
  0  success
  1  validation failed (only with --validate)
  2  parse error (malformed JSON on stdin or in --config file)
  3  I/O error (stdin/stdout/file)
```

The new subcommand-specific flags:

| Flag | Default | Notes |
|---|---|---|
| `--validate` | `false` | Run `types.ValidateRunConfig` on the result. Exit non-zero on failure. Without this, partial / chained configs are emitted as-is so they can be completed downstream. |
| `--compact` | `false` | Single-line JSON instead of indented. |
| `--redact` | `false` | Run `RunConfig.Redact()` before emit. |
| `--config <path>` | (none) | Base config from a file. Mutually exclusive with piped stdin (error if both are present). |

### Extended: `stirrup harness`

Two additions, both backwards-compatible:

1. **`--config -`** reads the base `RunConfig` from stdin. Also,
   when `--config` is not set *and* stdin is not a TTY, the harness
   auto-detects piped input and reads from stdin. (Explicit
   `--config -` is documented as the unambiguous form for scripted
   use; auto-detection exists for the pipe ergonomics shown in the
   examples above.)

2. **`--output-runconfig <path>`** writes the resolved, validated
   `RunConfig` as JSON to `<path>` (use `-` for stdout) and exits
   with code 0 without running the loop. Implicitly a dry-run. The
   path is created/overwritten with mode 0600; the parent must
   already exist. `--output-runconfig` and `--validate` always go
   together — if the resolved config is invalid the path is *not*
   written and the process exits non-zero with the validator's
   error.

The existing `--config <path>` continues to work unchanged. The
existing flag-only invocation continues to work unchanged.

### Worked examples

```sh
# Capture the config a flag-only invocation would have used:
stirrup run-config --mode execution --provider anthropic \
  --model claude-sonnet-4-6 --max-turns 50 > run.json

# Chained composition (the issue's canonical example):
stirrup run-config --model claude-opus-4-7 \
  | stirrup run-config --max-turns 100 \
  | stirrup run-config --mode execution \
  | stirrup harness --prompt "refactor module X"

# Starting from a checked-in config, override one field:
cat examples/runconfig/full.json \
  | stirrup run-config --model claude-haiku-4-5-20251001 \
  | stirrup harness --prompt "$(cat brief.txt)"

# Dry-run a flag-only invocation, capture what would have run:
stirrup harness --mode execution --provider gemini \
  --gcp-project my-proj --output-runconfig=resolved.json
# Process exits without running; resolved.json contains the config.

# Build a "save and replay" workflow for a specific configuration:
stirrup harness --output-runconfig=last-run.json --prompt "..."  # exits, captures config
stirrup harness --config last-run.json --prompt "rerun"          # uses captured config
```

### 1. Extract the shared builder (`harness/cmd/stirrup/cmd/runconfigbuilder.go`)

Refactor the two existing config-construction paths in
`harness/cmd/stirrup/cmd/harness.go` into one helper. The current
shape has duplication between `runHarness` Path 1 (`--config` +
`applyOverrides`) and Path 2 (`buildHarnessRunConfig` from flags
only), plus the WIF post-processing in `applyAnthropicWIFOverrides`
which already runs for both paths.

Introduce:

```go
// runconfigbuilder.go (new file, same package)

type RunConfigSources struct {
    Stdin       io.Reader   // nil means "do not read stdin"
    ConfigPath  string      // "" means no file; "-" means stdin
    Cmd         *cobra.Command
    Args        []string

    // Resolve controls late-stage mutation. The harness path uses
    // ResolveAll; the run-config path uses ResolveBase to keep
    // intermediate documents transformable.
    Resolve     ResolveMode
}

type ResolveMode int

const (
    ResolveBase     ResolveMode = iota // merge only, do not apply mode defaults, do not validate
    ResolveAll                         // apply mode defaults, run ValidateRunConfig
)

func BuildRunConfig(sources RunConfigSources) (*types.RunConfig, error)
```

Internals:

1. Determine the base: stdin if `ConfigPath == "-"` *or* (ConfigPath
   is empty and `Stdin` is non-nil and is not a TTY); else file if
   ConfigPath is set; else flag-only construction via
   `buildHarnessRunConfig`.
2. Validate "mutually exclusive base sources": `--config <path>`
   plus piped stdin is a hard error to avoid silent precedence
   surprises.
3. Apply the existing `applyOverrides(cmd, cfg, args)` — unchanged.
4. Apply the existing `applyAnthropicWIFOverrides` — unchanged.
5. If `Resolve == ResolveAll`: call `applyModeDefaults(cfg)`, then
   `types.ValidateRunConfig(cfg)`. Otherwise leave the document
   minimally mutated.
6. Resolve the prompt only when `Resolve == ResolveAll` (the
   existing `--prompt > positional > --prompt-file > STIRRUP_PROMPT >
   config.prompt` chain). For `ResolveBase`, leave the prompt at
   whatever the merge produced — possibly empty.

`runHarness` keeps its current outer shape but its two paths
collapse into one call to `BuildRunConfig` with `ResolveAll`. The
new `runRunConfig` calls `BuildRunConfig` with `ResolveBase` (or
`ResolveAll` when `--validate` is set).

The `runOptions` struct (CLI behaviour that doesn't fit on RunConfig)
stays in `runHarness` — it is not part of the wire shape and never
round-trips through `run-config`.

### 2. Add shared flag registration (`harness/cmd/stirrup/cmd/runconfigflags.go`)

Today every `RunConfig`-producing flag is registered inline in
`harnessCmd.init()`. Extract a helper:

```go
func addRunConfigFlags(cmd *cobra.Command) {
    // Identical body to the relevant portion of harnessCmd.init(),
    // covering every flag that maps to a RunConfig field. The
    // existing flag defaults are preserved verbatim.
}
```

`harnessCmd.init()` calls `addRunConfigFlags(harnessCmd)` and then
adds the four CLI-only behaviour flags it owns
(`--export-workspace-required`, `--output-runconfig`, plus the two
new subcommand-only flags below if applicable). The new
`runConfigCmd.init()` calls `addRunConfigFlags(runConfigCmd)` and
adds its three subcommand-specific flags (`--validate`, `--compact`,
`--redact`).

Validation of flag-only invariants (e.g.
`applyProviderRetryOverrides` rejecting sub-millisecond durations)
stays inside `BuildRunConfig` so both paths share it.

### 3. Implement `stirrup run-config` (`harness/cmd/stirrup/cmd/runconfig.go`)

```go
var runConfigCmd = &cobra.Command{
    Use:   "run-config [flags]",
    Short: "Emit a resolved RunConfig as JSON",
    Long:  `...`, // see "User-facing surface" above
    Args:  cobra.NoArgs,
    RunE:  runRunConfig,
}

func init() {
    rootCmd.AddCommand(runConfigCmd)
    addRunConfigFlags(runConfigCmd)
    runConfigCmd.Flags().Bool("validate", false, "...")
    runConfigCmd.Flags().Bool("compact", false, "...")
    runConfigCmd.Flags().Bool("redact", false, "...")
}

func runRunConfig(cmd *cobra.Command, args []string) error {
    resolve := ResolveBase
    if v, _ := cmd.Flags().GetBool("validate"); v {
        resolve = ResolveAll
    }
    cfg, err := BuildRunConfig(RunConfigSources{
        Stdin:      os.Stdin,
        ConfigPath: configPathFlag(cmd), // "" or path or "-"
        Cmd:        cmd,
        Args:       args,
        Resolve:    resolve,
    })
    if err != nil {
        return err // exits non-zero via cobra
    }
    if redact, _ := cmd.Flags().GetBool("redact"); redact {
        r := cfg.Redact()
        cfg = &r
    }
    return writeRunConfigJSON(os.Stdout, cfg, compactFlag(cmd))
}
```

Helper: `writeRunConfigJSON(w io.Writer, cfg *types.RunConfig, compact bool) error` uses `json.MarshalIndent` (or `json.Marshal` for compact) and writes plus a trailing newline.

Stdin handling notes:

- TTY detection uses `term.IsTerminal(int(os.Stdin.Fd()))` (already
  pulled in transitively, or add `golang.org/x/term` to the
  workspace if not). When stdin is a TTY, treat it as "no base
  config from stdin" and fall through to flag-only.
- The 1 MiB cap from `loadRunConfigFile`
  (`harness/cmd/stirrup/cmd/harness.go:597`) is reused for stdin
  via `io.LimitReader(os.Stdin, maxConfigFileBytes+1)`. Reading
  more triggers the same error message shape as the file path.
- `DisallowUnknownFields` is reused so typos in piped JSON fail
  loudly.

### 4. Wire `--config -` and auto-stdin into `stirrup harness`

In `loadRunConfigFile` (or a sibling `readRunConfigFrom(io.Reader)`
extracted from it), accept stdin when `path == "-"`. In
`runHarness`, when `configPath == ""` and stdin is non-TTY *and*
data is available, treat it as `--config -`.

The two rules together:

- `--config -` is the explicit, scripted opt-in.
- Bare stdin pipe is the ergonomic opt-in for interactive shells.
- `--config <path>` with non-TTY stdin is an error (ambiguous —
  reject early, mentioning both sources).

The existing stdio transport reads stdin for control events at run
time. After config is consumed (one-shot, bounded read at startup,
before transport init), stdin reaches EOF and the stdio transport
sees no further input — this matches the typical batch shape and
does not regress non-piped use. Document this in
`docs/configuration.md` under a "Reading from stdin" subsection.

### 5. Wire `--output-runconfig` into `stirrup harness`

In `runHarness`, after `BuildRunConfig` returns successfully (with
`ResolveAll`, so validation has run), check the flag:

```go
if outPath, _ := f.GetString("output-runconfig"); outPath != "" {
    return writeOutputRunConfig(outPath, cfg)
    // Process exits 0 without invoking runWithConfig.
}
```

`writeOutputRunConfig`:

- `"-"` writes to `os.Stdout`
- otherwise opens with `os.OpenFile(path, O_WRONLY|O_CREATE|O_TRUNC, 0600)`
- writes via the same `writeRunConfigJSON` helper used by
  `run-config`
- never writes on validation failure (the validator runs *before*
  this branch — confirm by control-flow walk)

Add the flag in `harnessCmd.init()`:

```go
f.String("output-runconfig", "", "Write the resolved RunConfig as JSON to <path> (use '-' for stdout) and exit without running. Useful for capturing the exact config a flag-only invocation would have used.")
```

Document that `--output-runconfig` does *not* skip the prompt
requirement: a prompt must still be resolvable, because the captured
config is "what would have run" — which includes the prompt.

### 6. Cross-field reasoning to preserve

The current resolution path has several cross-field rules that must
keep working unchanged after the refactor:

- `--api-key-ref` clearing on `--provider gemini` and on
  `--azure-tenant-id` (`harness/cmd/stirrup/cmd/harness.go:319`,
  `:880`, `:962`)
- `--gcp-credentials-file` implying `credential.type=gcp-service-account`
  (`:376`, `:925`)
- `--azure-tenant-id` implying `credential.type=azure-workload-identity`
  (`:401`, `:947`)
- `--anthropic-*` federation flags + env-var inference + apiKeyRef
  mutual exclusion (`applyAnthropicWIFOverrides`, `:1503`)
- `--permission-policy-file` implying `permissionPolicy.type=policy-engine`
  (`:413`, `:1036`)
- `--temperature` distinguishing absence from explicit 0 via
  `optionalFloat64Flag`
- Provider-retry duration → milliseconds conversion with
  sub-millisecond rejection (`retryDurationToMs`, `:556`)

After the refactor, every one of the cases above should be reachable
identically from the `run-config` path. The existing tests for these
cases (in `harness_test.go`) should continue to pass; the new tests
described below cover the path that exercises them via `run-config`.

### 7. Round-trip and idempotency

The pipeline use case requires that `stirrup run-config` produces
output that another `stirrup run-config` (with no overrides) can
parse and re-emit unchanged. Two invariants:

- **JSON validity**: the output passes `json.Unmarshal` with
  `DisallowUnknownFields` into `types.RunConfig`. Today this is
  trivially true because we marshal *from* the same struct.
- **Idempotency**: piping a config through `stirrup run-config`
  twice with no flags produces byte-identical output the second
  time. Defaults applied on the first pass do not need a second
  pass; mutations are idempotent.

Add a unit test that constructs a fully-populated `RunConfig`,
marshals it, feeds it back through `BuildRunConfig` with no flag
overrides, marshals again, and compares byte-for-byte.

## Documentation updates

- `docs/configuration.md`: add a "Building RunConfigs interactively"
  section describing the pipeline pattern, `--config -`, and
  `--output-runconfig`. Update the precedence table to mention
  stdin.
- `CLAUDE.md`: brief mention of the new subcommand in the "Where
  things live" table — `stirrup run-config` lives at
  `harness/cmd/stirrup/cmd/runconfig.go`.
- `AGENTS.md`: add the new subcommand to the CLI flags / commands
  section.
- `README.md`: one-paragraph mention of the pipeline pattern, with
  a single example.
- `examples/runconfig/README.md`: cross-link to the new subcommand
  ("To produce one of these from scratch, see `stirrup run-config
  --help`").

## Testing strategy

All new tests live under `harness/cmd/stirrup/cmd/` to match the
existing pattern. Build + lint must pass: `just` (default) and
`just lint`.

1. **`BuildRunConfig` unit tests** (`runconfigbuilder_test.go`):
   - flag-only → identical output to today's `buildHarnessRunConfig`
   - file-only via `--config tmp.json` → unchanged from today
   - stdin-only via `Stdin: bytes.NewReader(...)` and `ConfigPath: "-"`
   - stdin + flags → flag overrides applied
   - stdin + `--config <path>` → hard error, both sources cited
   - `ResolveBase` vs `ResolveAll` differ exactly in mode-defaults
     and validator invocation
   - `--prompt` resolution unchanged in `ResolveAll`; intentionally
     skipped in `ResolveBase`

2. **`runRunConfig` integration tests** (`runconfig_test.go`):
   - flag-only invocation emits JSON whose round-trip equals the
     flag-only path's resolved config (with mode defaults + validate
     applied for the comparison)
   - stdin → emit → re-pipe → emit produces byte-identical output
     (idempotency)
   - `--validate` rejects an invalid config with non-zero exit
   - `--compact` produces single-line JSON
   - `--redact` rewrites `secret://*` references but leaves all
     other fields intact
   - hostile inputs: 2 MiB stdin → size-cap rejection; malformed
     JSON → parse error with line/column from `json.SyntaxError`

3. **`--config -` and auto-stdin tests for harness** (extend
   `harness_test.go`):
   - `--config -` reads stdin
   - auto-stdin (non-TTY) reads stdin without `--config -`
   - `--config tmp.json` with piped stdin is rejected
   - `--config -` with a TTY stdin is rejected with a clear "no
     piped input" error

4. **`--output-runconfig` test** (extend `harness_test.go`):
   - `--output-runconfig=tmp.json` writes a parseable RunConfig and
     exits 0 without invoking the agent loop (verify by setting a
     stub `buildLoop` factory or by asserting no provider HTTP
     activity)
   - `--output-runconfig=-` writes to stdout
   - `--output-runconfig` with an invalid config exits non-zero and
     does *not* write the file
   - the captured config replays: piping it back into
     `stirrup harness --config -` (in a test harness, not a real
     run) produces the same resolved config

5. **Smoke test in CI**:
   - `stirrup run-config --mode execution --prompt "x" | jq -e .mode == "execution"`
   - Add as a one-liner under `scripts/` so it runs in the existing
     smoke suite per `docs/smoke-tests.md`.

## Edge cases / design decisions to call out in review

- **Auto-stdin vs explicit `--config -`.** Both are supported on the
  harness path; only auto is needed on `run-config` because that
  subcommand has no other use of stdin. The auto-detection uses
  `IsTerminal(stdin)`, not "stdin has data" (the latter requires a
  blocking read that defeats the purpose). This means
  `stirrup harness < /dev/null` is treated as "stdin piped, but
  empty" — fail loudly with "stdin is non-TTY but empty; pass
  `--config <path>` or remove the redirection" rather than silently
  falling through to flag-only.
- **Secret handling.** `secret://` references are *not* secrets;
  they point to env-var or external-store names. The `Redact()`
  method exists to scrub them for traces and recordings. By default,
  `run-config` emits references as-is so the output is replayable.
  `--redact` produces a "share-safe" copy. The `--output-runconfig`
  on `harness` emits *without* redaction because the captured
  config is intended to replay exactly — operators who need a
  share-safe artifact should pipe through `stirrup run-config --redact`.
- **`run-config` and validation.** Default is "do not validate" so
  partial / chained configs flow through the pipeline. `--validate`
  is the opt-in that turns the subcommand into a config-linter as
  well as an emitter. Validation runs `ValidateRunConfig`, which
  also mutates defaults (`applyCodeScannerDefault`,
  `applyProviderRetryDefaults`) — those mutations *are* visible in
  the emitted output. This is consistent with how the harness path
  treats validation today.
- **`run-config` and prompts.** The prompt is allowed to be empty
  on `run-config` output because the harness step at the end of the
  pipeline can provide it via `--prompt`, positional argument,
  `--prompt-file`, or `STIRRUP_PROMPT`. `--validate` on `run-config`
  does *not* enforce a non-empty prompt for the same reason —
  validation matches `ValidateRunConfig` exactly, which already
  accepts empty `Prompt` (the prompt-required check lives in
  `runHarness`, not in the validator).
- **`run-config` and CLI-only behaviour flags.** Flags that do not
  map to `RunConfig` fields (`--export-workspace-required`) are
  *not* exposed on `run-config` because they have nowhere to land
  in the emitted JSON. `--output-runconfig` is excluded from
  `run-config` for the same reason — that surface is its raison
  d'être.
- **Behaviour under `stirrup job`.** The control-plane-driven job
  entrypoint receives its `RunConfig` over gRPC and does not touch
  stdin. Neither new feature applies; documented in
  `docs/deployment.md` (no change required, but cross-check during
  doc review).
- **Stdio transport interaction.** Documented in
  `docs/configuration.md` that piping a config into
  `stirrup harness` consumes stdin at startup; subsequent stdio
  transport control-event input is therefore impossible for that
  run. This matches the batch shape the pipeline pattern targets.
  An operator who needs interactive control should use
  `--transport grpc`.

## Acceptance criteria

- [ ] `stirrup run-config --mode execution --provider anthropic`
      emits a valid JSON `RunConfig` to stdout and exits 0.
- [ ] `stirrup run-config --validate` exits non-zero on an invalid
      config (e.g. `--mode planning --executor api` with
      contradictory flags) with the validator's error on stderr.
- [ ] `cat config.json | stirrup run-config | stirrup run-config | diff - config.json` succeeds for any
      config in `examples/runconfig/` after one normalisation pass
      (the first pipe applies any default mutations; the second is
      idempotent).
- [ ] `stirrup run-config --model X | stirrup harness --prompt "Y"`
      runs the harness with model X and prompt Y, with no
      `--config` file involved.
- [ ] `stirrup harness --output-runconfig=out.json --prompt "x"`
      writes a parseable `RunConfig` to `out.json`, does not invoke
      the provider, and exits 0. Re-running with
      `stirrup harness --config out.json` produces an equivalent
      resolved config (the captured one).
- [ ] `stirrup harness --config tmp.json` and piped stdin together
      fail loudly, citing both sources.
- [ ] All existing `harness_test.go` tests pass unchanged.
- [ ] `just lint` (golangci-lint v2) passes.
- [ ] Docs updated per the list above; `doc-tone-reviewer` style
      maintained (impersonal voice; prefer "the harness" over
      "you").

## Files expected to change

```text
harness/cmd/stirrup/cmd/runconfigbuilder.go      [new]
harness/cmd/stirrup/cmd/runconfigbuilder_test.go [new]
harness/cmd/stirrup/cmd/runconfigflags.go        [new]
harness/cmd/stirrup/cmd/runconfig.go             [new]
harness/cmd/stirrup/cmd/runconfig_test.go        [new]
harness/cmd/stirrup/cmd/harness.go               [refactor: collapse Path 1/2 into BuildRunConfig; add --output-runconfig; add --config - and auto-stdin]
harness/cmd/stirrup/cmd/harness_test.go          [extend with the four new cases]
docs/configuration.md                            [pipeline pattern + precedence table update]
README.md                                        [one-paragraph mention]
AGENTS.md                                        [CLI commands list]
CLAUDE.md                                        [where-things-live row]
examples/runconfig/README.md                     [cross-link]
scripts/smoke-runconfig.sh                       [new — invoked by smoke suite]
```

No proto regen required: `RunConfig` schema is unchanged.

## References within the codebase

- Existing flag-only construction:
  `harness/cmd/stirrup/cmd/harness.go:260` (`buildHarnessRunConfig`)
- Existing `--config` + override path:
  `harness/cmd/stirrup/cmd/harness.go:807` (`applyOverrides`)
- Existing JSON loader with `DisallowUnknownFields` and size cap:
  `harness/cmd/stirrup/cmd/harness.go:597` (`loadRunConfigFile`)
- Cross-path post-processing:
  `harness/cmd/stirrup/cmd/harness.go:1503` (`applyAnthropicWIFOverrides`)
- Mode-derived defaults:
  `harness/cmd/stirrup/cmd/harness.go:576` (`applyModeDefaults`)
- Validator + default-applying mutations:
  `types/runconfig.go:1523` (`ValidateRunConfig`)
- `RunConfig` wire shape:
  `types/runconfig.go:89` and the matching proto in
  `proto/harness/v1/harness.proto`
- Redaction contract:
  `types/runconfig.go:414` (`RunConfig.Redact`)
- Existing examples:
  `examples/runconfig/full.json` (full surface), plus 12 others
  cross-linked from `docs/configuration.md`

## Why this design

- Reusing the existing `applyOverrides` precedence chain rather
  than inventing a new merge algorithm means the pipeline composes
  predictably with everything operators have already learned about
  `--config` + flags.
- A separate `run-config` subcommand (rather than only an
  `--output-runconfig` flag on harness) keeps the emit path free of
  prompt-required, secret-resolution, and transport concerns —
  exactly the friction the pipeline pattern is meant to eliminate.
- Auto-stdin on `harness` plus explicit `--config -` covers both
  the ergonomic-shell and scripted-pipeline use cases without
  forcing operators to choose one mental model.
- Keeping `--output-runconfig` as a harness-side dry-run (rather
  than "print and continue") means stdout stays clean for the
  agentic loop's normal output, and captured configs are
  unambiguously "what would have run" rather than entangled with a
  partial run's side effects.

rxbynerd:
Closes #240.

## Why

Capturing a fully-resolved `RunConfig` used to mean either
hand-crafting a JSON file or starting a real loop just to inspect
what the flags would have produced. This change adds the
`stirrup run-config` subcommand (emits the resolved document to
stdout without invoking the loop) and the `stirrup harness
--output-runconfig <path>` flag (dry-run capture of what *would* have
run). Together they enable a UNIX-style pipeline where each stage
layers one more adjustment before the final stage runs the agent,
and they give post-mortem replay a stable, file-based artefact. The
two surfaces converge on one resolver — `BuildRunConfig` in
`runconfigbuilder.go` — so the file-loading, stdin-piping,
flag-overriding, WIF-folding, prompt-chaining, and validation steps
are reachable from a single code path instead of the two parallel
branches that used to live in `runHarness`.

## What changed

- **New files**: `runconfigbuilder.go`, `runconfigflags.go`,
  `runconfig.go` (+ tests). `runconfigflags.go` extracts the shared
  flag registration so the two subcommands cannot drift.
- **`runHarness` refactor**: the pre-existing `--config`-with-overrides
  branch and flag-only branch collapse into one `BuildRunConfig`
  call. New `--output-runconfig` and `--config -` / auto-stdin
  handling.
- **`buildHarnessRunConfig` refactor**: split into a `Core` helper
  (no `applyModeDefaults`) and the public wrapper (calls
  `applyModeDefaults` for backward compatibility). `BuildRunConfig`
  uses the core helper so its `ResolveBase` path can emit a
  minimally-mutated document.
- **Docs**: `docs/configuration.md` gains a "Building RunConfigs
  interactively" section; README.md gets a paragraph; AGENTS.md
  picks up the new flag rows and a `run-config` flag table;
  CLAUDE.md's where-things-live table points at the new files;
  `examples/runconfig/README.md` cross-links to the subcommand.
- **Smoke**: `scripts/smoke-runconfig.sh` exercises the four
  offline acceptance criteria (emitter shape, idempotency,
  `--validate`, `--output-runconfig`).

## Deviation from the spec

The spec calls for auto-stdin detection via `IsTerminal(stdin)` —
any non-TTY stdin is treated as a piped base. In practice that
trips every existing `harness_test.go` fixture because the
`go test` runner inherits stdin from the terminal as a non-TTY
character device. The implementation narrows the auto-detection to
named pipes (shell `|`) and regular-file redirects (`<
config.json`); other non-TTY shapes (char devices including
`< /dev/null`, Unix sockets) fall through to flag-only construction.
The explicit `--config -` form continues to enforce the spec's
"non-TTY but empty" failure mode on EOF input, which keeps the
acceptance criterion about `< /dev/null` reachable without trapping
the test suite. The rationale and the cost (operators who
deliberately pipe `< /dev/null` get the prompt-required error rather
than a "stdin was empty" error) is documented in the
`isStdinPiped` comment.

## Test plan

- [x] `just` (build + test) passes — 26 new tests added under
  `harness/cmd/stirrup/cmd/`.
- [x] `just lint` passes (golangci-lint v2).
- [x] Manual smoke: `./scripts/smoke-runconfig.sh` passes
  end-to-end after `just build`.
- [x] Pipeline pattern works end-to-end:
  `stirrup run-config --model X | stirrup run-config --mode execution | stirrup harness --output-runconfig=-`.
- [x] `stirrup harness --config /tmp/x` plus piped stdin fails
  loudly with both sources cited.
- [x] `stirrup harness --output-runconfig=out.json --prompt "x"`
  writes a parseable `RunConfig` and exits 0; re-running with
  `--config out.json` produces an equivalent resolved config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,26 @@ OPENAI_KEY="$(op read "op://Private/qeu6gafabhkpsm6hhzattx6p4m/credential")" ./s
 
 ```
 
+### Composing configs in a pipeline
+
+For development workflows that build up a `RunConfig` incrementally,
+`stirrup run-config` emits the resolved JSON document without
+invoking the loop, and `stirrup harness` reads a base config from
+stdin so each stage layers one more adjustment before the final stage
+runs the agent:
+
+```sh
+stirrup run-config --model claude-opus-4-7 \
+  | stirrup run-config --max-turns 100 \
+  | stirrup run-config --mode execution --executor container \
+  | stirrup harness --prompt "refactor module X"
+```
+
+`stirrup harness --output-runconfig <path>` captures the exact
+config a flag-only invocation *would* have used — useful for
+post-mortem replays or pinning a stable configuration. See
+[`docs/configuration.md`](docs/configuration.md#building-runconfigs-interactively).
+
 ### In GitHub Actions
 
 See [`.github/workflows/smoke-anthropic.yml`](.github/workflows/smoke-anthropic.yml) for an example of using `stirrup harness` in a GitHub Actions workflow via [Anthropic Workload Identity Federation](https://platform.claude.com/docs/en/manage-claude/workload-identity-federation).
--- a/scripts/smoke-runconfig.sh
+++ b/scripts/smoke-runconfig.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+# smoke-runconfig.sh — verify `stirrup run-config` emits a parseable
+# RunConfig and the canonical pipeline shape round-trips idempotently.
+#
+# Four checks run sequentially:
+#
+#   1. Flag-only invocation produces JSON with the operator-set mode
+#      and prompt visible. Asserted with `jq -e` so a structural drift
+#      in the emitted JSON trips this script before it lands in
+#      a release.
+#   2. Idempotency: piping a config through `stirrup run-config` twice
+#      produces byte-identical output. The first pass normalises;
+#      the second is a no-op.
+#   3. --validate exits non-zero on a contradictory config.
+#   4. `stirrup harness --output-runconfig <path>` writes a
+#      parseable RunConfig and exits cleanly without invoking the
+#      provider (no ANTHROPIC_API_KEY set; the dry-run branch must
+#      short-circuit before any provider HTTP).
+#
+# Run from the repo root after `just build`:
+#   ./scripts/smoke-runconfig.sh
+#
+# Honours STIRRUP_BIN as an override path (e.g. for CI builds that
+# stage the binary somewhere other than ./stirrup).
+
+set -euo pipefail
+
+STIRRUP_BIN="${STIRRUP_BIN:-./stirrup}"
+TMP_DIR="$(mktemp -d)"
+trap 'rm -rf "${TMP_DIR}"' EXIT
+
+if [[ ! -x "${STIRRUP_BIN}" ]]; then
+    echo "FAIL: stirrup binary not found at ${STIRRUP_BIN}" >&2
+    echo "  Run \`just build\` first or set STIRRUP_BIN to the binary path." >&2
+    exit 1
+fi
+
+# Check 1: flag-only invocation produces parseable JSON with the
+# operator-set mode visible.
+mode_out=$(${STIRRUP_BIN} run-config --mode execution --prompt "x" \
+    | jq -er '.mode')
+if [[ "${mode_out}" != "execution" ]]; then
+    echo "FAIL: --mode execution did not surface in emitted JSON (got: ${mode_out})" >&2
+    exit 1
+fi
+echo "ok: run-config emits parseable JSON with operator-set mode"
+
+# Check 2: idempotency. Pipe a config through twice and diff the two
+# outputs. The first run may apply default mutations (e.g. cobra
+# defaults); the second must not move the document any further.
+${STIRRUP_BIN} run-config --mode execution --prompt "x" > "${TMP_DIR}/pass1.json"
+${STIRRUP_BIN} run-config < "${TMP_DIR}/pass1.json" > "${TMP_DIR}/pass2.json"
+if ! diff -q "${TMP_DIR}/pass1.json" "${TMP_DIR}/pass2.json" >/dev/null; then
+    echo "FAIL: run-config is not idempotent" >&2
+    diff "${TMP_DIR}/pass1.json" "${TMP_DIR}/pass2.json" >&2 || true
+    exit 1
+fi
+echo "ok: run-config | run-config is idempotent"
+
+# Check 3: --validate rejects a contradictory config. Bedrock + the
+# CLI-default `claude-sonnet-4-6` is the issue #65 trap the validator
+# rejects with an inference-profile remediation hint.
+if ${STIRRUP_BIN} run-config --validate \
+    --mode execution \
+    --provider bedrock \
+    --prompt "x" >/dev/null 2>&1; then
+    echo "FAIL: --validate accepted bedrock + sonnet-4-6 alias" >&2
+    exit 1
+fi
+echo "ok: --validate rejects invalid configs"
+
+# Check 4: --output-runconfig writes a parseable file and exits
+# cleanly without invoking the provider. We rely on the dry-run
+# branch firing before any HTTP would happen; the absence of
+# ANTHROPIC_API_KEY here proves the loop did not start.
+${STIRRUP_BIN} harness --output-runconfig "${TMP_DIR}/captured.json" \
+    --prompt "x" --mode planning
+captured_mode=$(jq -er '.mode' < "${TMP_DIR}/captured.json")
+if [[ "${captured_mode}" != "planning" ]]; then
+    echo "FAIL: --output-runconfig did not capture --mode planning (got: ${captured_mode})" >&2
+    exit 1
+fi
+echo "ok: --output-runconfig captures the resolved RunConfig"
+
+echo "all run-config smoke checks passed"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,6 +81,21 @@ Requires `ANTHROPIC_API_KEY` environment variable for the default Anthropic prov
 | `--transport-addr` | (none) | gRPC target address, required when transport is grpc |
 | `--followup-grace` | `0` | Seconds to keep gRPC open for follow-ups (env: `STIRRUP_FOLLOWUP_GRACE`) |
 | `--log-level` | `info` | Log level: debug, info, warn, error |
+| `--config` | (none) | JSON `RunConfig` path. Use `-` for stdin; auto-detected on a non-TTY pipe. |
+| `--output-runconfig` | (none) | Write the resolved `RunConfig` JSON to `<path>` (or `-` for stdout) and exit without running. Dry-run capture for replay or post-mortem. |
+
+### CLI Flags (`stirrup run-config`)
+
+Emits a fully-resolved `RunConfig` JSON document without invoking the
+loop. Accepts every `RunConfig`-producing flag from `stirrup harness`
+plus a base via stdin or `--config <path>`. See
+[`docs/configuration.md`](docs/configuration.md#building-runconfigs-interactively).
+
+| Flag | Default | Description |
+|---|---|---|
+| `--validate` | `false` | Run `types.ValidateRunConfig` and exit non-zero on failure. |
+| `--compact` | `false` | Single-line JSON instead of indented (2-space). |
+| `--redact` | `false` | Apply `RunConfig.Redact()` before emit (rewrites `secret://` references). |
 
 ## Architecture
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -138,7 +138,10 @@ Quick map for "I need to change X" lookups:
 | Result sink (RunResult payload) | `harness/internal/resultsink/` |
 | Workspace GCS export | `harness/internal/workspaceexport/` |
 | `RunConfig` validation | `types/runconfig.go` |
-| CLI flag definitions | `harness/cmd/stirrup/cmd/harness.go` |
+| CLI flag definitions (shared between `harness` and `run-config`) | `harness/cmd/stirrup/cmd/runconfigflags.go` |
+| CLI flag definitions (`harness`-only behaviour: `--export-workspace-required`, `--output-runconfig`) | `harness/cmd/stirrup/cmd/harness.go` |
+| Resolved-config builder (file/stdin/flags → RunConfig) | `harness/cmd/stirrup/cmd/runconfigbuilder.go` |
+| `stirrup run-config` subcommand | `harness/cmd/stirrup/cmd/runconfig.go` |
 | gRPC schema | `proto/harness/v1/harness.proto` (then `buf generate`) |
 | Eval suite parser | `eval/spec/` |
 | Eval CLI subcommands | `eval/cmd/eval/main.go` |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -21,20 +21,83 @@ fails fast with a clear error rather than being silently dropped.
 
 ## Precedence
 
-When both `--config` and explicit flags are passed, the order of
-precedence is:
+When `--config`, piped stdin, or explicit flags are passed, the order
+of precedence is:
 
-1. **File** — `--config` populates the full `RunConfig`.
+1. **Base config** — either `--config <path>` (file) or stdin
+   (`--config -`, or auto-detected on a non-TTY pipe). Mutually
+   exclusive: passing both is a hard error so the operator never
+   has to guess which source won.
 2. **Explicit flags** — flags whose `cmd.Flags().Changed(...)` bit is
-   set replace the corresponding file-provided field.
+   set replace the corresponding base field.
 3. **Defaults** — flags left at their default value do **not**
-   override the file. This is what makes `--config` ergonomic: you
-   do not have to clear every default to keep the file's intent.
+   override the base. This is what makes `--config` ergonomic:
+   defaults can stay defaults while the file's intent is preserved.
 
 The positional `prompt` argument is a fallback only. It fills the
-prompt slot when the file omits it and `--prompt` is not set, but
-neither the file's `prompt` nor an explicit `--prompt` is overridden
+prompt slot when the base omits it and `--prompt` is not set, but
+neither the base's `prompt` nor an explicit `--prompt` is overridden
 by a positional.
+
+## Building RunConfigs interactively
+
+For development workflows that compose a `RunConfig` incrementally —
+or just to capture what a flag-only invocation *would* have run — the
+harness ships two surfaces that complement `--config <path>`:
+
+- `stirrup run-config` emits a fully-resolved `RunConfig` JSON
+  document to stdout without invoking the agentic loop. Every
+  RunConfig-producing flag from `stirrup harness` is honoured; a
+  base config can arrive via stdin (the pipeline pattern) or
+  `--config <path>`.
+- `stirrup harness --output-runconfig <path>` writes the resolved
+  RunConfig as JSON to `<path>` (use `-` for stdout) and exits
+  without running. The captured document is exactly what would have
+  been handed to the loop, so it can be checked into source control
+  for replay or compared with `diff` between runs.
+
+The two surfaces together support a UNIX-style pipeline where each
+stage layers one more adjustment before the final stage runs the
+agent:
+
+```sh
+stirrup run-config --model claude-opus-4-7 \
+  | stirrup run-config --max-turns 100 \
+  | stirrup run-config --mode execution --executor container \
+  | stirrup harness --prompt "refactor X"
+```
+
+`stirrup run-config` exposes three subcommand-specific flags:
+
+| Flag | Default | Notes |
+|---|---|---|
+| `--validate` | `false` | Run `types.ValidateRunConfig` on the resolved document and exit non-zero on failure. Without this flag, partial / chained configs are emitted as-is so a downstream stage can complete them. |
+| `--compact` | `false` | Emit single-line JSON instead of indented (2-space). |
+| `--redact` | `false` | Apply `RunConfig.Redact()` before emit. Rewrites `secret://` references to `secret://[REDACTED]`. The result is share-safe but no longer runnable as-is — operators who need a runnable replay should omit the flag. |
+
+`stirrup harness --output-runconfig` always emits the unredacted form
+because its purpose is exact replay; pipe through
+`stirrup run-config --redact` when a share-safe artifact is wanted.
+
+### Reading from stdin
+
+`stirrup harness` accepts a base `RunConfig` from stdin in two shapes:
+
+- **`--config -`** is the explicit, scripted opt-in. It always
+  reads stdin and fails loudly if stdin is a terminal or carries no
+  data.
+- **Auto-detection on a non-TTY pipe** triggers when `--config` is
+  unset and stdin is a named pipe (`|`) or a regular-file
+  redirection (`< config.json`). Other non-TTY shapes — including
+  `< /dev/null` and the stdin handed to tests by `go test` — fall
+  through to flag-only construction so the harness does not trap
+  noninteractive automation.
+
+Piping a config into `stirrup harness` consumes stdin at startup,
+before transport initialisation, so the stdio transport sees EOF for
+any subsequent control-event input. This matches the batch shape the
+pipeline pattern targets — operators who need interactive control
+should use `--transport grpc`.
 
 ## CLI flags
 
--- a/examples/runconfig/README.md
+++ b/examples/runconfig/README.md
@@ -42,20 +42,12 @@ fails fast with a clear error rather than being silently dropped.
 
 ## Precedence
 
-When the user passes both `--config <path>` and individual flags, the
-order of precedence is:
-
-1. **File** — `--config` populates the full `RunConfig`.
-2. **Explicit flags** — flags whose `cmd.Flags().Changed(...)` bit is
-   set replace the corresponding file-provided field.
-3. **Defaults** — flags left at their default value do **not**
-   override the file. This is what makes `--config` ergonomic:
-   defaults can stay defaults while the file's intent is preserved.
-
-The positional `prompt` argument is a fallback only. It fills the
-prompt slot when the file omits it and `--prompt` is not set, but
-neither the file's `prompt` nor an explicit `--prompt` is overridden
-by a positional.
+The canonical resolution table — covering `--config <path>`, piped
+stdin (`--config -` or auto-detected on a non-TTY pipe), explicit
+flag overrides, and the positional prompt fallback — lives in
+[`docs/configuration.md`](../../docs/configuration.md#building-runconfigs-interactively).
+The same section documents the `stirrup run-config | stirrup harness`
+pipeline pattern these examples plug into.
 
 ## Annotated example walkthrough
 
--- a/examples/runconfig/README.md
+++ b/examples/runconfig/README.md
@@ -3,6 +3,13 @@
 This directory contains example `RunConfig` JSON files for the
 `stirrup harness --config <path>` flag.
 
+To produce one of these from scratch — either from a flag-only
+invocation or by composing several `run-config` stages — see
+`stirrup run-config --help` and the pipeline pattern documented in
+[`docs/configuration.md`](../../docs/configuration.md#building-runconfigs-interactively).
+The same file format is consumed by `--config` and emitted by
+`run-config`, so a captured config round-trips into a replay run.
+
 ## Source of truth
 
 The canonical schema for these files is the protobuf definition at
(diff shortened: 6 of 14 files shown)


rxbynerd:
## Wave 1 remediation pass

Addressed all 3 MUST-FIX items and 8 of 9 SHOULD-FIX items from the consolidated review. SF-8 was a no-op (the comment typos described in the brief don't exist in the current code — both lines 722 and 1178 already start with valid `//`).

**Commit range:** `c29ccbd..aa282b1` (5 commits on top of the existing 3)

### MUST-FIX
- **MF-1** (`c29ccbd`): Removed the duplicate `applyAnthropicWIFOverrides` call from inside `applyOverrides`. The post-override call in `BuildRunConfig` is now the single site. New `TestBuildRunConfig_AnthropicWIFWarnFiresOnce` pins warn-once behaviour via captured `slog` output.
- **MF-2** (`d4f0bed`): Added four `TestRunHarness_Stdin*` integration tests covering `--config -`, auto-stdin detection, ambiguous `--config <path>` + piped stdin, and `--config -` with no pipe. Tests drive `runHarness` end-to-end via `os.Pipe` stdin swap + `--output-runconfig=-` capture (shared helpers `withPipedStdin` / `captureStdout`).
- **MF-3** (`c687da7`): Refactored `runRunConfig` around a `runRunConfigWithIO(cmd, args, stdin, stdout)` seam. `runRunConfigForTest` now calls through the real entry point; new `TestRunRunConfigWithIO_TableDriven` exercises every cobra flag branch (flag-only, `--compact`, `--redact`, `--validate`). `runRunConfig` coverage went from 0% to 100%.

### SHOULD-FIX
- **SF-1, SF-3** (`aa282b1`): `writeOutputRunConfig` now surfaces the `Close` error via a `writeAndCloseRunConfig` testable seam. New `TestWriteOutputRunConfig_CloseError` (close-error and write-error precedence subtests) + `TestRunHarness_OutputRunConfigBadPathErrors` (OpenFile error branch).
- **SF-2** (`aa282b1`): `TestRunHarness_ConfigPathFollowupGraceFromEnv` rewritten to call `BuildRunConfig(ResolveAll)` rather than `loadRunConfigFile` + an inline env-var copy. The test now actually exercises the production branch.
- **SF-4, SF-5, SF-6** (`2cfd045`): Added `TestIsStdinPiped_*` (named pipe, regular file, Stat error, char device — pins the documented `go test` deviation), `TestWriteRunConfigJSON_WriteError` (both Write call sites via `failWriter`), and `TestBuildRunConfig_ApplyOverridesError` / `TestBuildRunConfig_WIFOverridesError` (error-propagation arms previously bypassed).
- **SF-7** (`c687da7`): `newTestHarnessCommand` now delegates to `addRunConfigFlags` and only registers the two harness-only behaviour flags (`--export-workspace-required`, `--output-runconfig`) itself. A flag added to the shared registry is picked up by both factories automatically.
- **SF-8**: No-op. Both lines flagged in the brief (`harness.go:722` and `harness.go:1178`) already begin with valid `//`. `gofmt -l harness.go` returns nothing.
- **SF-9** (`aa282b1`): `examples/runconfig/README.md` precedence section replaced with a forward reference to `docs/configuration.md#building-runconfigs-interactively`, the canonical table that already documents the stdin source.

### Verification
- `just` (build + test) green
- `just lint` (golangci-lint v2) clean, 0 issues
- `./scripts/smoke-runconfig.sh` all four checks pass
- Test count in `harness/cmd/stirrup/cmd/` went from 188 → 212 (+24)

scripts/smoke-runconfig.sh
rxbynerd, line 1: follow-up: port to just script

README.md
rxbynerd, line 1: we should avoid editing the readme, especially when we're linking to subdocuments: unroll manually later, investigate why this recurs.

PR #254, opened 2026-05-22
cli: add STIRRUP_CONFIG env var for default --config path

### LINKED ISSUE #241: CLI: add STIRRUP_CONFIG env var for default --config path
## Background

`stirrup harness` accepts a base `RunConfig` from `--config <path>`. Operators who use one config across many invocations currently retype the flag every time, or wrap the binary in a shell alias. A `STIRRUP_CONFIG` environment variable would mirror `KUBECONFIG` / `AWS_CONFIG_FILE` and let the path be set once per shell session.

Related: this composes naturally with the work in #240 (run-config subcommand and stdin support).

## Request

Add a single env-var fallback for the `--config` flag, with the resolution order:

1. `--config <path>` flag (explicit)
2. `--config -` or piped stdin (per #240)
3. `STIRRUP_CONFIG` env var
4. (no config — flag-only construction)

The env var follows the same shape as the flag:

- A filesystem path: loaded the same way `loadRunConfigFile` already loads `--config`.
- The literal value `-`: treated as `--config -` (stdin). Useful for orchestration scripts that pipe in a config but don't want to inject a flag.

Reject the env var when piped stdin is also present (same rule as `--config <path>` vs piped stdin in #240): both naming a base source is ambiguous.

## Implementation notes

- The single place that needs the change is the source-resolution step inside the shared `BuildRunConfig` helper proposed in #240. If #240 has not landed yet, the change is inside `runHarness` at the `configPath, _ := f.GetString("config")` site (`harness/cmd/stirrup/cmd/harness.go:1228`).
- Log a single `slog.Debug` line when the env var is the chosen source so operators can audit precedence at debug log level.
- Document the env var in `docs/configuration.md` under the "Precedence" section.

## Acceptance criteria

- [ ] `STIRRUP_CONFIG=/path/to/config.json stirrup harness --prompt "x"` runs with the env-var config.
- [ ] `STIRRUP_CONFIG=/path/to/config.json stirrup harness --config /other.json --prompt "x"` uses `--config /other.json` (explicit flag wins).
- [ ] `STIRRUP_CONFIG=- cat config.json | stirrup harness --prompt "x"` reads from stdin.
- [ ] `STIRRUP_CONFIG=/path cat config.json | stirrup harness --prompt "x"` exits non-zero with a clear "both env var and piped stdin specify a base config" error.
- [ ] Documented in `docs/configuration.md` precedence table and prose.
- [ ] Existing tests pass.

## Files expected to change

```text
harness/cmd/stirrup/cmd/harness.go              (or runconfigbuilder.go from #240)
harness/cmd/stirrup/cmd/harness_test.go         (new env-var test cases)
docs/configuration.md                           (precedence section)
```

rxbynerd:
Closes #241

## Summary

Adds `STIRRUP_CONFIG` as a fourth-rank fallback in the `BuildRunConfig` precedence chain, mirroring `KUBECONFIG` / `AWS_CONFIG_FILE`. Operators who use one RunConfig across many invocations can set it once per shell session instead of retyping `--config` or wrapping the binary in an alias. An explicit `--config` always wins over the env var, so an ad-hoc override is still a one-flag affair; combining `STIRRUP_CONFIG=<path>` with piped stdin is rejected the same way `--config <path>` + piped stdin already is, with an error that cites the env var by name so the source of the failure is unambiguous.

The change layers cleanly into the #240 precedence work: the entire env-var read lives inside `loadBaseRunConfig` so all four base-source decisions stay in one switch statement rather than splitting across `runHarness` and the helper.

A single `slog.Debug` line fires when the env var is the chosen source so operators can audit precedence at debug log level.

## Base branch

This PR targets `feat/issue-240-runconfig-generator`, not `main`. PR #251 (the #240 PR) is open but unmerged; targeting it as the base keeps the diff scoped to only #241's changes. GitHub will auto-retarget this PR to `main` when #251 merges.

## Test plan

- [x] `STIRRUP_CONFIG=/path stirrup harness --output-runconfig -` emits the env-var config
- [x] `STIRRUP_CONFIG=/path stirrup harness --config /other.json --output-runconfig -` emits `/other.json` (flag wins)
- [x] `STIRRUP_CONFIG=- stirrup harness --output-runconfig - < cfg.json` reads from stdin
- [x] `STIRRUP_CONFIG=/path stirrup harness --output-runconfig - < cfg.json` exits 1 with `ambiguous config sources: both env var STIRRUP_CONFIG="..." and piped stdin specify a base config`
- [x] `just test` (`go test ./harness/... ./types/... ./eval/...`) green
- [x] `golangci-lint run ./harness/cmd/stirrup/cmd/...` clean
- [x] Six new test cases in `runconfigbuilder_test.go` cover env-var supplies-base, explicit-flag-wins, dash-opts-into-stdin, path+stdin-ambiguous, dash-requires-stdin, empty-ignored, and debug-logged paths

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -21,23 +21,59 @@ fails fast with a clear error rather than being silently dropped.
 
 ## Precedence
 
-When `--config`, piped stdin, or explicit flags are passed, the order
-of precedence is:
+When `--config`, piped stdin, `STIRRUP_CONFIG`, or explicit flags
+are passed, the order of precedence is:
 
-1. **Base config** — either `--config <path>` (file) or stdin
-   (`--config -`, or auto-detected on a non-TTY pipe). Mutually
-   exclusive: passing both is a hard error so the operator never
-   has to guess which source won.
+1. **Base config** — one of the following, in descending priority:
+
+   | Rank | Source | Notes |
+   |---|---|---|
+   | 1 | `--config <path>` flag | Explicit. Wins outright. |
+   | 2 | `--config -` flag | Reads the base from piped stdin. |
+   | 3 | Auto-detected piped stdin | Triggered when `--config` is absent and stdin is a pipe or a redirected regular file (not a TTY). |
+   | 4 | `STIRRUP_CONFIG` env var | Filesystem path, or the literal `-` to opt into stdin. Consulted only when `--config` is absent. |
+
+   The four sources are mutually exclusive. Combining a path-shaped
+   source (`--config <path>` or `STIRRUP_CONFIG=<path>`) with piped
+   stdin is a hard error so operators never have to guess which
+   source won. `STIRRUP_CONFIG=-` with piped stdin is allowed because
+   the env var is opting into the stdin path, not naming a separate
+   base.
 2. **Explicit flags** — flags whose `cmd.Flags().Changed(...)` bit is
    set replace the corresponding base field.
 3. **Defaults** — flags left at their default value do **not**
    override the base. This is what makes `--config` ergonomic:
    defaults can stay defaults while the file's intent is preserved.
 
+When the env var is the chosen source, the harness emits a single
+`slog.Debug` line naming the path (or `-`) so operators can audit
+precedence at debug log level without grepping the source.
+
 The positional `prompt` argument is a fallback only. It fills the
 prompt slot when the base omits it and `--prompt` is not set, but
 neither the base's `prompt` nor an explicit `--prompt` is overridden
 by a positional.
+
+### `STIRRUP_CONFIG` ergonomics
+
+`STIRRUP_CONFIG` mirrors `KUBECONFIG` and `AWS_CONFIG_FILE`: setting
+it once per shell session removes the need to retype `--config` on
+every invocation, while leaving the flag itself available for ad-hoc
+overrides.
+
+```sh
+export STIRRUP_CONFIG="$HOME/.config/stirrup/default.json"
+
+# Uses the env-var config.
+stirrup harness --prompt "explain this repo"
+
+# Explicit --config wins over the env var for this one invocation.
+stirrup harness --config ./experimental.json --prompt "try the new model"
+
+# Pipe a config in and ignore the env var path; STIRRUP_CONFIG=- opts
+# into stdin without naming a separate source.
+STIRRUP_CONFIG=- some-tool emit-config | stirrup harness --prompt "x"
+```
 
 ## Building RunConfigs interactively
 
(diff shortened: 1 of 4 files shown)


rxbynerd:
## Remediation cycle: 7aa387c..0a8a9f6

Acted on the 1 MUST-FIX and all 5 SHOULD-FIX items from the consolidated review. DEFERRED / DECLINED items left untouched as instructed.

### MUST-FIX
- **B1** (`runconfigbuilder.go:184-211`) — moved both `slog.Debug` calls out of the env-var resolution block and into the switch arms (`case filePath != "":` and `case explicitStdin:` success path), so the chosen-source trace only fires when the source is actually consumed. Pinned by extending `TestBuildRunConfig_EnvVarPathPlusStdinIsAmbiguous` to capture slog output and assert no `STIRRUP_CONFIG` trace on the ambiguity path.

### SHOULD-FIX
- **S1** (`runconfigbuilder.go:213-217`) — wrapped `loadRunConfigFile` error with `"STIRRUP_CONFIG: %w"` when env-sourced, so a bad env-var path is attributed to the env var instead of looking like a flag error. Test: `TestBuildRunConfig_EnvVarPathNotFound`.
- **S2** (`runconfigbuilder.go:205-211`) — split the dash-without-stdin error into two messages: `"no stdin reader was provided"` when `sources.Stdin == nil`, `"stdin is a terminal"` when a TTY is attached. Test: tightened `TestBuildRunConfig_EnvVarDashRequiresStdin` to assert the new message and the absence of `"terminal"`.
- **S3** (`runconfigbuilder.go:203`) — dash branch now uses `slog.Debug("...", "path", "-")` matching the path branch's structured field. Test: new `TestBuildRunConfig_EnvVarDashDebugLogged` plus `path=` assertion added to `TestBuildRunConfig_EnvVarDebugLogged`.
- **S4** (`harness_test.go:4710-4782`) — added `TestRunHarness_EnvVarConfigLoadsBase`, an integration test driving `runHarness → BuildRunConfig → os.Getenv("STIRRUP_CONFIG")` end-to-end using the `withPipedStdin` / `captureStdout` pattern from PR #251.
- **S5** (`runconfigbuilder.go:184`) — chose to **trim** with `strings.TrimSpace`. Reason: shell env-var editors and CI secret stores routinely smuggle in a trailing newline; `open  /path: no such file` with a phantom leading space is a worse failure than treating the trimmed value as authoritative. Decision pinned by `TestBuildRunConfig_EnvVarLeadingWhitespaceTrimmed`.

### Verification
- `just` — all packages pass.
- `just lint` — 0 issues.
- Test count: 183 → 187 (+4 new tests in `cmd/`).

### Not addressed (per brief)
- D1 stale-comment update at `harness.go:399-405` — carry as a separate one-liner.
- D2 / N4 doc-table precedence rank inversion — defer to doc-only pass.
- D3 `slog.SetDefault` race documentation — safe today; no parallel tests.

PR #258, opened 2026-05-22
cli: add stirrup trace subcommand family

### LINKED ISSUE #244: CLI: stirrup trace subcommand family (show/tail/stats/grep)
## Background

Stirrup emits JSONL traces (`traceEmitter.type=jsonl`) capturing every turn, tool call, token usage delta, permission decision, and security event. Operators inspect these today via `cat`, `jq`, and ad-hoc scripts. The eval framework already has a JSONL trace parser (`eval/runner/replay.go`); the main binary should expose first-party inspection tooling so operators do not reinvent it.

## Request

Add a `stirrup trace` subcommand family that operates on a JSONL trace file (or stdin):

| Subcommand | Behaviour |
|---|---|
| `stirrup trace show <file>` | Pretty-prints the trace with chronological ordering, tool call inputs/outputs unfolded, ANSI colour when stdout is a TTY. Default selector: all events. |
| `stirrup trace tail [-f] <file>` | Streams new records as they are appended (`tail -f` semantics). Useful while a long run is in progress. |
| `stirrup trace stats <file>` | Aggregates: total turns, total tokens (in/out/cache hit), tool-call count by tool, longest turn duration, total wall-clock, top-N slowest tool calls, permission denials, security events. JSON output for machine consumption (`--output=json`) and text for humans. |
| `stirrup trace grep <pattern> <file>` | Filter events by JSON path (`--jq '.type == "tool_call"'`) or substring (`<pattern>`). Emits matching records. Removes the dependency on having `jq` installed. |

All subcommands accept `-` as the file argument to read from stdin (composes with `--trace-emitter=jsonl --trace=/dev/stdout` once that lands, or with `tail -F | stirrup trace …`).

## Implementation notes

- The trace event types are defined in `types/runtrace.go` and `types/messages.go`. The parser is in `eval/runner/replay.go` — extract it into a shared package (`types/trace/reader.go`?) so both the eval framework and the new `trace` subcommand consume it.
- `trace show` should colourise by event type: tool calls in blue, errors in red, permission decisions in yellow, etc. Disable colour automatically when stdout is not a TTY (`golang.org/x/term`).
- `trace stats` should be cheap (single-pass, no rebuild of state). Mirror metric names used in `harness/internal/observability/` so the same vocabulary covers OTel metrics and offline stats.
- `trace tail -f` uses `os.File.Seek` + a polling loop. Don't pull in `fsnotify` for this — the trace files are append-only and a 100 ms poll is fine.
- This subcommand lives on the main `stirrup` binary, NOT `stirrup-eval`, so operators inspecting a production run do not need to install the eval binary.

## Edge cases

- Truncated or in-progress JSONL: skip malformed lines with a `slog.Debug` log; do not fail the command. Operators tail-watching a live run get partial output.
- Compressed traces (`.gz`)? Defer to a follow-up — operators can `zcat foo.jsonl.gz | stirrup trace show -`.
- Schema drift: include the trace's `RunConfig.runId` and harness version in `stats` output so operators correlate traces across versions.

## Acceptance criteria

- [ ] `stirrup trace show example.jsonl` produces a readable, chronological dump.
- [ ] `stirrup trace stats example.jsonl --output=json | jq .totalTurns` returns the turn count.
- [ ] `stirrup trace tail -f live.jsonl` streams appended lines.
- [ ] `stirrup trace grep --jq '.type == "permission_denied"' live.jsonl` filters by JSON path.
- [ ] Reading from stdin works for every subcommand (`-` as the file arg).
- [ ] Colour disabled on non-TTY stdout; controllable via `--color=auto|always|never`.
- [ ] Documented in a new `docs/trace-inspection.md` page and cross-linked from `docs/observability-cloud.md`.

## Files expected to change

```text
harness/cmd/stirrup/cmd/trace.go                (new — subcommand root)
harness/cmd/stirrup/cmd/trace_show.go           (new)
harness/cmd/stirrup/cmd/trace_tail.go           (new)
harness/cmd/stirrup/cmd/trace_stats.go          (new)
harness/cmd/stirrup/cmd/trace_grep.go           (new)
types/trace/reader.go                            (new — extracted from eval/runner/replay.go)
eval/runner/replay.go                            (refactor to consume types/trace/reader.go)
docs/trace-inspection.md                         (new)
docs/observability-cloud.md                      (cross-link)
```

rxbynerd:
## Summary

Closes #244.

Operators inspecting `traceEmitter.type=jsonl` output have been reaching for `cat`, `jq`, and ad-hoc shell. This PR exposes four first-party `stirrup trace` subcommands so the same workflow runs without installing the eval binary or shelling `jq`:

- `trace show` — chronological pretty-print with TTY colour autodetection (`--color=auto|always|never`).
- `trace tail` — polling follow-tail (`-f`, configurable `--interval`); no fsnotify or other notification dep.
- `trace stats` — aggregate token/tool-call/outcome counters in text or `--output=json`, metric names mirror the OTel vocabulary in `harness/internal/observability/metrics.go`.
- `trace grep` — substring + a tiny `--jq` predicate (`<path> ( == | != | contains ) <value>`) so the workflow does not require a `jq` binary.

Every subcommand accepts `-` as the file argument to read from stdin (pipes with `tail -F`, `gunzip -c`, `gcloud storage cat`, etc.). Malformed and oversized lines are skipped with `slog.Debug` so a tail-watch of an in-flight run keeps streaming partial output.

The JSONL parser is lifted into a new zero-dep `types/trace` package and shared with the existing eval runner (`eval/runner/runner.go::parseTraceFile`), so the new subcommands and the eval framework speak one implementation.

## Test plan

- [x] `just` (build + full test suite) passes
- [x] `just lint` passes
- [x] Existing `eval/runner` tests still pass (`TestReplayRecording_*`, `TestParseTraceFile`)
- [x] New tests cover the reader (malformed/oversized lines, follow-tail, stdin) and every subcommand path (show colour modes, stats JSON+text, grep substring/jq/array-path/invert, tail one-shot + follow)
- [x] Smoke-tested the binary end-to-end against a sample JSONL trace (show/stats/grep/tail one-shot)

### DIFF
--- a/docs/observability-cloud.md
+++ b/docs/observability-cloud.md
@@ -11,6 +11,10 @@ model — what spans, metrics, and resource attributes Stirrup emits —
 see [`safety-rings.md`](safety-rings.md) and the
 "OpenTelemetry trace emitter" section of the repo-root `CLAUDE.md`.
 
+For operators inspecting the JSONL trace emitter's output (offline
+pretty-print, follow-tail, aggregate stats, JSON-path filtering),
+see [`trace-inspection.md`](trace-inspection.md).
+
 ## Quick choice
 
 | Where do you want telemetry to land? | Use |
--- a/docs/trace-inspection.md
+++ b/docs/trace-inspection.md
@@ -78,6 +78,14 @@ SIGINT (Ctrl-C) terminates the follow loop cleanly. `tail -f -` reads
 from stdin — the follow flag is redundant in that mode because stdin
 already blocks until more bytes are written.
 
+> **Truncation and rotation are not followed.** `tail -f` keeps the
+> same file handle open. If the file is truncated or rotated out
+> from under it, the offset stays past the new EOF and the follow
+> loop appears to stall with no further output and no error.
+> Restart the command to pick up the rotated file. Full
+> `tail --follow=name` semantics (inode-tracking) are out of scope
+> — stirrup's own traces are append-only.
+
 ## `trace stats`
 
 ```sh
--- a/docs/trace-inspection.md
+++ b/docs/trace-inspection.md
@@ -1,0 +1,254 @@
+# Inspecting JSONL trace files
+
+A run configured with `traceEmitter.type=jsonl` writes a newline-delimited
+JSON document describing the run's turns, tool calls, token usage,
+verification results, and final outcome. The `stirrup trace`
+subcommand family inspects those files first-party so operators do
+not have to reinvent the wheel with `cat`, `jq`, and ad-hoc shell.
+
+This page is the operator walkthrough. For the wire schema, see the
+`types.RunTrace`, `types.TurnTrace`, and `types.ToolCallSummary`
+definitions in [`types/runtrace.go`](../types/runtrace.go).
+
+## Quick choice
+
+| Want to… | Use |
+|---|---|
+| Read a finished trace end-to-end | `stirrup trace show <file>` |
+| Watch a still-running trace | `stirrup trace tail -f <file>` |
+| Pull aggregate counters into a dashboard | `stirrup trace stats <file> --output=json` |
+| Pull every record matching a predicate | `stirrup trace grep --jq '<path> <op> <value>' <file>` |
+
+Every subcommand accepts `-` as the file argument to read from stdin,
+so the family composes with shell pipelines:
+
+```sh
+tail -F live.jsonl | stirrup trace show -
+gunzip -c archived.jsonl.gz | stirrup trace stats - --output=json
+```
+
+A `.gz` archive is read by piping through `gunzip` or `zcat`; the
+subcommands themselves are uncompressed-only and do not embed a
+gzip dependency.
+
+## `trace show`
+
+```sh
+stirrup trace show example.jsonl
+```
+
+Pretty-prints the trace in chronological order: the run metadata
+header, every tool call in the order it ran, the verification
+verdicts, and a per-tool aggregate at the foot. Tool calls flagged
+unsuccessful are surfaced in red; the outcome is colour-coded green
+(success), red (error/verification_failed/tool_failures),
+yellow (max_turns / max_tokens / budget_exceeded / stalled / timeout),
+grey (cancelled), or blue (anything unrecognised).
+
+ANSI colour is auto-detected: stdout to a TTY is coloured; stdout
+piped to a file or `less` is not. Override the detection with:
+
+```sh
+stirrup trace show example.jsonl --color=never   # plain text always
+stirrup trace show example.jsonl --color=always  # colour even when piping
+```
+
+## `trace tail`
+
+```sh
+stirrup trace tail live.jsonl       # one-shot equivalent of show
+stirrup trace tail -f live.jsonl    # follow the file as it grows
+```
+
+`tail` reads the file once and prints every record (one-shot). With
+`-f` (or `--follow`), it polls the file at `--interval` (default 100ms)
+and prints records as they are appended. The polling interval is
+configurable for very quiet files:
+
+```sh
+stirrup trace tail -f live.jsonl --interval=500ms
+```
+
+`tail` uses a polling loop rather than a kernel notification API.
+Trace files are append-only and a sub-second poll is sufficient to
+keep an operator's terminal in sync without inflating the dependency
+footprint.
+
+SIGINT (Ctrl-C) terminates the follow loop cleanly. `tail -f -` reads
+from stdin — the follow flag is redundant in that mode because stdin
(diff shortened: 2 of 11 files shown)


rxbynerd:
## Remediation cycle 1 — done

Pushed 5 new commits on top of the existing branch (no rewrite, no force push):

| SHA | Subject | Addresses |
|---|---|---|
| `688bec0` | `cli: reset trace grep scanner past oversized lines` | **M1** (silent drop after oversized line), **S2** (magic buffer constants) |
| `66ff93a` | `cli: honour NO_COLOR and complete --color values` | **M2** (NO_COLOR), **M3** (parseColorMode coverage), **S1** (shell completion) |
| `354640c` | `cli: cover trace grep != and stdin paths for show/tail/stats` | **M4** (`!=` coverage), **M5** (stdin for show/tail/stats) |
| `5a239f2` | `cli: source trace stats RunID from RunConfig and widen test coverage` | **S4** (RunID precedence), **S5** (outcomeColor), **S6** (parseJQValue), **S7** (walkPath), **S8** (jsonEqual float/bool) |
| `dd6b5f5` | `docs: warn that trace tail does not follow truncation or rotation` | **S3** (truncation/rotation doc) |

### MUST-FIX status

- [x] **M1** — `runTraceGrepWith` resets the scanner on `bufio.ErrTooLong` and continues, matching `Reader.Next()` recovery. Test pins both pre-cap and post-cap records appear in output.
- [x] **M2** — `shouldColor` checks `os.Getenv("NO_COLOR")` before the TTY probe in the auto path. `colorAlways` deliberately still wins (explicit opt-in not overridden by ambient env).
- [x] **M3** — `TestParseColorMode` table covers `auto`, `always`, `force`, `yes`, `never`, `no`, `off`, empty, and an invalid case (`rainbow`) that exercises the error branch.
- [x] **M4** — `TestRunTraceGrepWith_JQNotEqual` exercises both the compile and eval arms of `!=`.
- [x] **M5** — `TestRunTraceShowWith_StdinPath`, `TestRunTraceStatsWith_StdinPath`, `TestRunTraceTailWith_StdinPath` cover the missing three.

### SHOULD-FIX status

All eight SHOULD-FIX items landed:

- **S1** — `addColorFlag` now calls `RegisterFlagCompletionFunc("color", …)` so `auto|always|never` surfaces in bash/zsh tab completion.
- **S2** — Buffer caps in grep now reference `tracereader.MaxLineBytes` (and `MaxLineBytes/16` for initial) so the two paths cannot drift.
- **S3** — `Tail` godoc and `docs/trace-inspection.md` both call out that truncation/rotation silently stalls the follow loop, with `tail --follow=name` explicitly out of scope.
- **S4** — `TraceStats.absorb` prefers `t.Config.RunID` and falls back to `t.ID`.
- **S5** — `TestOutcomeColor` table covers every arm: success/error/yellow-bucket/cancelled/default.
- **S6** — `TestParseJQValue` table covers string/true/false/null/numeric/error branches.
- **S7** — `TestWalkPath` covers object/array-indexed/non-numeric-on-array/OOB/scalar-default/missing.
- **S8** — `TestRunTraceGrepWith_JQNumericAndBool` exercises `jsonEqual` float64 and bool arms.

None skipped.

### DEFER status

- **D1** — Filed [issue #260](https://github.com/rxbynerd/stirrup/issues/260) (`trace: add permission-denial counter to RunTrace and trace stats`). Requires a schema change to `RunTrace` and is genuinely out of scope here.
- **D2** — Implicitly tracked alongside D1 (version-injection seam for tests). Not separately filed; flag if a discrete issue is wanted.
- **D3** — Tail-stdin direct test gap is now covered by `TestRunTraceTailWith_StdinPath` (`-` path); the brief noted this as low-risk and it folded naturally into the M5 work.
- **D4** — Posted [a clarifying comment on #244](https://github.com/rxbynerd/stirrup/issues/244#issuecomment-4518873812) noting the spec referenced `eval/runner/replay.go` but the parser was in `runner.go`, and that the PR correctly retargeted.

### Validation

- `just` (build + full test suite): green
- `just lint` (golangci-lint v2): `0 issues.`
- Targeted suites (`go test ./types/trace/... ./harness/cmd/stirrup/cmd/... ./eval/runner/...`): all pass
- Trace-test count: 15 → 27 (+12)

### Notes

- The conflict between M1 and S2 (S2 noted it was "superseded if M1 is addressed by refactoring grep to use `tracereader.Reader.Next()` directly") was resolved by taking the safer fix: keep the dedicated scanner (so substring matching still operates on the raw line bytes without an intermediate marshal), but reference `tracereader.MaxLineBytes` to satisfy S2's drift concern. Both reviewers' intents land.

Issue #277, opened 2026-05-22
v0.1: demonstrable evals — parent tracking issue


rxbynerd:
## Goal

Ship v0.1 of stirrup with **stirrup-eval demonstrably driving stirrup's development**. The end-to-end loop runs locally on the OSS FileStore lakehouse — no managed-store dependency, no control plane.

The demonstrable feature, in one sentence: *operators run stirrup on a real codebase, traces of those runs feed `mine-failures`, the mined regression suite gates future PRs against the CI baseline, and a meaningful pass-rate signal connects lab runs to production reality.*

The exit criterion is **#13 — first mined suite from real PRs**, running against this repo's own development sessions. Every building block below exists to make #13 produce a representative, falsifiable suite rather than a stage prop.

## Demo narrative

1. `stirrup harness --trace tmp/sessions/*.jsonl` captures real coding-agent sessions on this repo (dogfood; the harness building its own features).
2. `stirrup-eval ingest --trace tmp/sessions/*.jsonl --lakehouse var/lakehouse` populates both `traces/` and `recordings/`.
3. `stirrup-eval baseline --lakehouse var/lakehouse` shows current pass / fail / inconclusive rates against real work — not termination state, *quality* state.
4. `stirrup-eval mine-failures --lakehouse var/lakehouse --outcome failed --accept-quarantine --output eval/suites/mined.hcl` turns a week of failures into a regression suite, each task carrying the failing-turn context.
5. `stirrup-eval run --suite eval/suites/mined.hcl --output results/ --concurrency 8` runs the mined suite at real cadence.
6. `stirrup-eval compare --current results/result.json --baseline eval/baselines/mined.json` blocks the PR that regressed.
7. `stirrup-eval drift --lakehouse var/lakehouse --window 7d` flags when production quality moves out from under the eval suite.
8. `stirrup-eval replay --suite eval/suites/mined.hcl --recording <id>` iterates on judges without re-burning tokens.

Every step above either works today and gets significantly better, or doesn't work today and needs the work tracked below.

## Building blocks

Coordinator agents implementing v0.1 should treat the checklist as authoritative. Tick boxes as each building block lands.

### Phase 1 — recording emitter (gates most of the rest)

[...]

PR #278, opened 2026-05-22
trace: stream full transcript as JSONL events

### LINKED ISSUE #270: trace: stream full transcript as JSONL events (split from #248)
## Background

Today the JSONL trace emitter (`harness/internal/trace/jsonl.go`) buffers `TurnTrace` and `ToolCallTrace` summaries during a run, then emits a single `RunTrace` line at `Finish`. The provider and executor already produce the full transcript — `ModelInput.Messages`, `ModelOutput`, and `ToolCallRecord` with `Input` (raw JSON) and `Output` (string) — but the emitter discards everything except sizes.

The consequence is that `eval/runner/replay.go::ReplayRecording`, `harness/internal/provider/replay.go::ReplayProvider`, and `harness/internal/executor/replay.go::ReplayExecutor` have no production-derived corpus to operate on. `mine-failures` queries `QueryRecordings` and gets nothing back, because nothing in the production harness writes recordings. The eval framework's reproducibility story is structurally blocked on this missing writer.

This issue is the recording-writer half of #248, split out so it can land independently of the harder `--continue` resume work. #248 retains the resume/continuation scope.

## Scope

Replace (or augment) the JSONL trace emitter so it streams typed events as the run progresses, instead of buffering and emitting one summary line at `Finish`. The new file format is line-delimited JSON where each line carries a `kind` discriminator:

```jsonl
{"kind":"run_started","runId":"...","config":{...redacted...},"startedAt":"..."}
{"kind":"turn_record","turn":1,"modelInput":{...},"modelOutput":[...],"toolCalls":[...]}
{"kind":"tool_call_record","turn":1,"name":"read_file","input":{...},"output":"..."}
{"kind":"turn_record","turn":2,"modelInput":{...},...}
{"kind":"run_finished","trace":{...RunTrace summary...},"completedAt":"..."}
```

A successful run produces one `run_started`, N `turn_record` lines (each carrying its own `tool_call_record` entries inline), and one `run_finished`. An interrupted run (SIGKILL, crash, OOM) may end without `run_finished`; the on-disk JSONL is still parseable up to the last completed turn.

### Required behaviour

- **Backward-compatible read path.** `types/trace/reader.go` must continue to load existing single-blob `RunTrace` JSONL traces emitted before this change. They parse as a `run_finished` event only, with no recording attached.
- **Defence-in-depth scrubbing.** `LogScrubber` runs over `modelOutput`, tool call `input`, and tool call `output` before the line is written. The redaction guarantee matches the current `RunConfig.Redact()` posture for traces.
- **No raw secret material.** Tool inputs/outputs and message content go through the existing scrubber. Provider configs are redacted via `RunConfig.Redact()` in the `run_started` event.
- **Append-only durability.** Lines are flushed after each write so a SIGKILL leaves a parseable partial recording on disk.
- **Wire-shape parity.** Event kinds map cleanly to future gRPC `HarnessEvent` variants so the same emitter can serve the control-plane wire later.

### Out of scope

- `stirrup harness --continue <recording>` — restoration of conversation state into a fresh process. That stays in #248.
- gRPC streaming of recording events to a control plane. Reuses this wire shape later; not part of this issue.
- Compression. Recordings can be megabytes; compression is a follow-up if it becomes a real problem on disk.
- Schema versioning across major bumps. A `schemaVersion` field on `run_started` is fine; full version negotiation is post-v0.1.

## Acceptance

- [ ] `harness/internal/trace/` carries a streaming emitter producing the event-typed JSONL above.
- [ ] `types/trace/reader.go` reassembles a `RunRecording` from a complete event stream and a `RunTrace` from the terminating `run_finished` event.
- [ ] Existing single-blob `RunTrace` traces still parse (as a `run_finished`-only stream).
- [ ] A SIGKILL'd run produces a partial JSONL file that the reader can parse up to the last complete event.
- [ ] `LogScrubber` is applied to `modelOutput`, tool call `input`, and tool call `output` at emit time. Tested.
- [ ] No measurable regression in run wall time for the existing eval suites (sanity check; large diff in recording-write cost would imply a buffering issue).
- [ ] `docs/eval.md` and `docs/trace-inspection.md` updated to describe the new on-disk shape.

## References

- `harness/internal/trace/jsonl.go` — current emitter.
- `types/runtrace.go::RunRecording`, `TurnRecord`, `ToolCallRecord` — target shapes.
- `types/trace/reader.go` — read-side reassembly lives here.
- `harness/internal/provider/replay.go`, `harness/internal/executor/replay.go`, `eval/runner/replay.go` — consumers that need recordings.
- #248 — sibling issue retaining the `--continue` scope.
- #267 — atomic `writeJSON` (relevant for the ingest side that consumes this).

rxbynerd:
## Summary

Replaces the buffered single-blob JSONL trace emitter with a streaming, event-typed line-delimited JSON writer. Each agentic-loop turn now lands a `turn_record` on disk the moment it completes, carrying the full `ModelInput.Messages`, the model's `ModelOutput`, and every tool call's raw input/output. A `SIGKILL`'d run leaves a parseable JSONL file up to the last completed event.

Defence-in-depth scrubbing (`harness/internal/security.Scrub`) runs over all transcript text and raw tool I/O before any line is written, giving recordings a second independent redaction layer on top of the existing `RunConfig.Redact()` posture.

`types/trace.Reader` transparently accepts both the legacy single-blob shape (one `RunTrace` per line) and the new streaming shape; `Reader.Next` still yields one `*RunTrace` per completed run so the eval runner's `parseTraceFile` keeps working unchanged. A new `Reader.ReadRecording` walks the stream and reassembles a `*types.RunRecording` for replay / mine-failures.

Closes #270.
Part of #277.

## Commits

- `trace: stream typed events from JSONLTraceEmitter` — interface + emitter
- `trace: reader handles streaming events and reassembles RunRecording` — read path
- `core: wire full-turn transcripts through TraceEmitter` — loop / dispatch wiring
- `docs: describe streamed-events JSONL trace shape` — `eval.md` + `trace-inspection.md`

## Test plan

- [x] `just` passes (build + tests, 36 packages)
- [x] `just lint` passes (golangci-lint v2 clean)
- [x] New tests pin: full lifecycle, session-name round trip, scrubbing, partial-stream (SIGKILL), legacy/streaming equivalence, recording reassembly (streaming + partial + legacy), forward compatibility for unknown event kinds
- [ ] Operator smoke: run a real harness session with `--trace tmp/x.jsonl`, confirm a) at least one `turn_record` event per turn, b) the file is parseable by `stirrup trace show` mid-run, c) `Reader.ReadRecording` reassembles the run
- [ ] Performance: no measurable wall-time regression on existing eval suites (~50 extra one-line writes per 10-turn run; well under measurement noise)

### DIFF
--- a/docs/eval.md
+++ b/docs/eval.md
@@ -284,6 +284,23 @@ traces, and let the `replay` runner re-evaluate old recordings under
 new judge criteria without re-running the harness
 (`eval/runner/replay.go`).
 
+#### Where recordings come from
+
+`TurnRecord` payloads are produced live by the
+`traceEmitter.type=jsonl` emitter in
+`harness/internal/trace/jsonl.go`. The emitter streams one
+`turn_record` event per agentic-loop turn into the configured trace
+file, carrying the full `ModelInput.Messages`, the model's
+`ModelOutput` content blocks, and every tool call's raw
+`Input`/`Output`. `types/trace.Reader.ReadRecording` reassembles a
+`types.RunRecording` from the event stream, including from
+partially-written files left behind by an interrupted run.
+
+Pre-streaming traces (single-blob `RunTrace` lines) parse through the
+same reader as recordings with no transcript turns; replay against
+those files is a degenerate case that succeeds only if the eval suite
+asks for zero turns.
+
 ### Trace lakehouse
 
 The `TraceLakehouse` interface (`types/lakehouse.go`) abstracts
--- a/docs/trace-inspection.md
+++ b/docs/trace-inspection.md
@@ -7,8 +7,39 @@ subcommand family inspects those files first-party so operators do
 not have to reinvent the wheel with `cat`, `jq`, and ad-hoc shell.
 
 This page is the operator walkthrough. For the wire schema, see the
-`types.RunTrace`, `types.TurnTrace`, and `types.ToolCallSummary`
-definitions in [`types/runtrace.go`](../types/runtrace.go).
+`types.RunTrace`, `types.TurnTrace`, `types.TurnRecord`, and
+`types.ToolCallSummary` definitions in
+[`types/runtrace.go`](../types/runtrace.go).
+
+## On-disk shape
+
+A JSONL trace file is a stream of one-line JSON events, each carrying
+a `kind` discriminator. A complete run produces:
+
+```jsonl
+{"kind":"run_started","schemaVersion":"1","runId":"...","config":{...redacted...},"startedAt":"..."}
+{"kind":"turn_record","turn":1,"modelInput":{...},"modelOutput":[...],"toolCalls":[...]}
+{"kind":"tool_call_record","turn":1,"name":"read_file","input":{...},"output":"..."}
+{"kind":"turn_record","turn":2,"modelInput":{...},"modelOutput":[...]}
+{"kind":"run_finished","trace":{...RunTrace summary...},"completedAt":"..."}
+```
+
+`turn_record` carries the full transcript the model saw and produced
+that turn, including raw tool inputs and outputs. `tool_call_record`
+is emitted in addition so a live consumer sees each call as soon as
+it lands. `run_finished` carries the same `RunTrace` summary the
+legacy single-blob format used; `stirrup trace show / stats / grep`
+read it transparently regardless of which on-disk shape produced it.
+
+Lines are flushed after each write, so a `SIGKILL`'d run leaves a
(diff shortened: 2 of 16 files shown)


rxbynerd:
review notes:

- docs will need holistic pass later, ideally by stirrup itself
- `harness/internal/trace/gcs.go` gcs should be destination for a full trace, rather than legacy trace, therefore should consume turn records.

PR #283, opened 2026-05-23
lakehouse: narrow TraceLakehouse to read surface

### LINKED ISSUE #109: lakehouse: narrow TraceLakehouse interface to read surface; resolve HarnessEvent.trace
## Background

Two related cleanups originally identified during #104, refined during v0.1 planning (#277).

### 1. `HarnessEvent.trace` (proto/harness/v1/harness.proto:122)

Documented as carrying a `RunTrace` on `done` events, but `loop.go:294-298` emits `done` without populating it (`grpc_translate.go:26-28` would forward a non-nil trace; the loop never produces one). The wire field is dead code today.

Two options:

- **Populate it.** The trace would carry the simplified 7-field proto shape (`run_id`, `turns`, `tokens`, `cost`, `duration`, `stop_reason`). Requires the loop to wire `Trace.Finish` output into `event.Trace` before `Emit`. Useful as a CP "I have all the metrics now" signal even when OTel is the canonical wire.
- **Drop it in proto v2.** Rely on OTel + a control-plane recording upload to carry trace data. Cleanest, but requires a proto major version cut.

Either is acceptable; the decision can land independently of #2 below.

### 2. `TraceLakehouse` interface narrowing

`TraceLakehouse.StoreTrace` and `StoreRecording` (`types/lakehouse.go:9-27`) are never called outside test code in the production harness. `grep -rn "StoreTrace\|StoreRecording" .` returns only the interface, the `FileStore` implementation, test callers, and (after PR #268) the `stirrup-eval ingest` CLI.

The original #104 follow-up suggested deleting these methods entirely. **That's wrong** — it would break `stirrup-eval ingest`, which uses them as the persistence call site after parsing JSONL trace input.

The correct narrowing:

- **`TraceLakehouse` interface** keeps only the read surface (`QueryTraces`, `QueryRecordings`, `Metrics`, `Close`) — that is what the eval CLI's read-side subcommands (`baseline`, `mine-failures`, `drift`, `compare-to-production`, `replay`) actually consume.
- **`FileStore` keeps `StoreTrace` and `StoreRecording` as concrete methods**, not part of the interface. The `stirrup-eval ingest` CLI constructs `lakehouse.NewFileStore` directly (already does today) and calls these methods on the concrete type, not via the interface.
- **No cloud-backed adapter implements `StoreTrace`/`StoreRecording`** by design — production writers will be the control plane, not the harness or the CLI.

This preserves the architectural intent (the interface describes what consumers depend on; production writers don't go through it) without breaking the local OSS path that v0.1 demonstrable evals (#277) depends on.

## Scope

- [ ] Narrow `TraceLakehouse` (`types/lakehouse.go`) to `QueryTraces`, `QueryRecordings`, `Metrics`, `Close`.
- [ ] Move `StoreTrace` and `StoreRecording` to be `FileStore`-concrete methods (no interface).
- [ ] Update `stirrup-eval ingest` (`eval/cmd/eval/main.go`) to keep using the concrete `*FileStore` type (already does — change is just removing the interface reference if any was added in PR #268).
- [ ] Verify no remaining production caller depends on `TraceLakehouse.StoreTrace` or `StoreRecording` (only tests and `cmdIngest`).
- [ ] Separate sub-task: decide `HarnessEvent.trace` populate-or-drop. ADR or direct PR; not a blocker for the interface change.

## Coordination

- v0.1 milestone (#277). Should land **before** the ingest extension (#271) so the ingest call site has a clean post-narrowing target.
- Compatible with #270 (streamed transcript emitter) — that work writes JSONL to disk and doesn't touch the lakehouse interface; the ingest side (#271) is where the write call sites live.

## References

- `types/lakehouse.go:9-27` — current `TraceLakehouse` interface.
- `eval/lakehouse/filestore.go` — `FileStore` implementation.
- `eval/cmd/eval/main.go` (PR #268's `cmdIngest`) — the legitimate write call site that must keep working.
- `proto/harness/v1/harness.proto:122` — `HarnessEvent.trace` dead field.
- `docs/research/issue-104/FOLLOWUPS.md` (FU-2) — original framing.

rxbynerd:
## Summary

`StoreTrace` and `StoreRecording` move off the `TraceLakehouse` interface and live as concrete methods on `*FileStore`. The interface now contains only the read surface that eval consumers depend on (`QueryTraces`, `QueryRecordings`, `Metrics`, `Close`). A compile-time interface-conformance assertion in `filestore.go` pins the contract.

Architectural effect: future cloud-backed adapters are not required to implement write methods; production writes flow through the control plane; the OSS / dev path keeps writing via the concrete `*FileStore`.

Closes #109 (the interface-narrowing half — `HarnessEvent.trace` populate-or-drop is deferred per the issue body).
Part of #277.
Stacked on #282.

## Test plan

- [x] `just` passes
- [x] `just lint` passes
- [x] Compile-time `var _ types.TraceLakehouse = (*FileStore)(nil)` pins the contract
- [x] Existing FileStore tests cover both surfaces (read methods via the interface, write methods on the concrete type)

### DIFF
--- a/eval/lakehouse/filestore.go
+++ b/eval/lakehouse/filestore.go
@@ -1,4 +1,10 @@
-// Package lakehouse provides file-based implementations of the TraceLakehouse interface.
+// Package lakehouse provides a file-based implementation of the read
+// surface of types.TraceLakehouse plus the concrete write methods
+// (StoreTrace, StoreRecording) that stirrup-eval ingest calls
+// directly. The write methods are intentionally NOT part of the
+// TraceLakehouse interface — see #109 for the architectural rationale
+// (production writes flow through the control plane; cloud-backed
+// adapters never implement write).
 package lakehouse
 
 import (
@@ -18,12 +24,24 @@ const (
 	recordingsDir = "recordings"
 )
 
-// FileStore implements types.TraceLakehouse backed by JSON files on disk.
-// Traces are stored in <root>/traces/<id>.json and recordings in
-// <root>/recordings/<runId>.json.
+// FileStore implements the read surface of types.TraceLakehouse
+// (QueryTraces, QueryRecordings, Metrics, Close) backed by JSON
+// files on disk, and additionally exposes the concrete StoreTrace
+// and StoreRecording methods that `stirrup-eval ingest` constructs
+// a *FileStore for. Traces are stored in <root>/traces/<id>.json
+// and recordings in <root>/recordings/<runId>.json.
+//
+// A compile-time assertion below pins that *FileStore satisfies the
+// narrowed types.TraceLakehouse contract; if a method is ever
+// removed from the read surface accidentally, the build fails.
 type FileStore struct {
 	rootDir string
 }
+
+// _ var pins the interface conformance contract. Any change to
+// types.TraceLakehouse that removes a method from *FileStore will
+// break the build here before it can ship.
+var _ types.TraceLakehouse = (*FileStore)(nil)
 
 // NewFileStore creates a FileStore rooted at rootDir, creating the necessary
 // subdirectories if they don't already exist.
--- a/types/lakehouse.go
+++ b/types/lakehouse.go
@@ -2,17 +2,19 @@ package types
 
 import "context"
 
-// TraceLakehouse abstracts storage and querying of production run data.
-// Implementations may back this with files, Postgres JSONB, BigQuery, or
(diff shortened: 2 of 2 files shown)


rxbynerd:
Lakehouse interface should be over gRPC, therefore control plane implementations already consuming stirrup protos can provide a lakehouse for evals.

PR #284, opened 2026-05-23
eval: ingest streamed recording events into FileStore

### LINKED ISSUE #271: eval: ingest streamed recording events into FileStore
## Background

PR #268 (closing #266) adds `stirrup-eval ingest` for single-blob `RunTrace` JSONL files: each line is a complete trace, ingest writes one file per trace into `<lakehouse>/traces/`. This issue extends ingest to consume the streamed event format introduced by the trace-emitter rewrite (sibling issue — recording emitter splits out of #248).

After this lands, the chain `stirrup harness --trace ... && stirrup-eval ingest --trace ... --lakehouse ...` populates both `traces/` and `recordings/` from real production runs, making `mine-failures`, `eval replay`, and the `ReplayProvider`/`ReplayExecutor` codepaths usable end-to-end on local hardware.

## Scope

Extend `stirrup-eval ingest` to handle the typed-event JSONL produced by the new trace emitter.

### Behaviour

- **Format detection.** First non-blank line is inspected. If it has a `kind` discriminator, treat as event-stream; otherwise fall back to the existing per-line `RunTrace` decoder. Single ingest invocation may mix both formats across multiple `--trace` arguments.
- **Reassembly.** Buffer events per `runId` until `run_finished` is observed, then commit one `RunTrace` to `traces/` and one `RunRecording` to `recordings/`. Use `FileStore.StoreTrace` and `FileStore.StoreRecording` directly — see #109 for the interface narrowing context.
- **Partial recordings.** A stream that ends without `run_finished` (interrupted run) gets ingested with a synthesised `FinalOutcome.Outcome = "interrupted"` and the `Turns` collected so far. Default behaviour. Operators wanting to skip partials can pass `--skip-partial`.
- **Idempotent.** Re-ingesting the same JSONL produces identical on-disk state (last-write-wins, same as today).
- **Validation.** Each `tool_call_record` event must have a corresponding parent `turn_record`. Out-of-order or orphan events produce a per-line stderr warning; the rest of the stream still ingests.
- **Per-line scrubbing.** Defence in depth: `LogScrubber` runs on the ingest side too, so a trace produced by an older binary without scrubbing still gets scrubbed on the way into the lakehouse.

### Out of scope

- gRPC ingest from a live control plane (that's the control-plane writer, not the CLI).
- Schema migration for traces produced by future breaking emitter changes.

## Acceptance

- [ ] `stirrup-eval ingest` correctly reassembles `RunRecording` from a streamed JSONL trace, writing both `traces/<id>.json` and `recordings/<id>.json`.
- [ ] Existing single-blob JSONL traces still ingest (no `recordings/` entry produced for them — there's nothing to reassemble).
- [ ] Interrupted streams ingest as partial recordings by default; `--skip-partial` excludes them.
- [ ] Path-traversal hardening on `runId` matches the existing rules introduced in PR #268.
- [ ] Tests cover: a full streamed trace; a single-blob legacy trace; a partial stream; mixed-format invocation; an out-of-order event.
- [ ] `docs/eval.md` ingest section documents the new format support.

## Dependencies

- Sibling issue: recording emitter (Phase 1 of v0.1).
- #267 — atomic `writeJSON`. Recording payloads make non-atomic writes a real corruption risk under concurrent ingest; this needs to land first or in parallel.
- #109 — `TraceLakehouse` interface narrowing. The `StoreTrace`/`StoreRecording` methods used here become FileStore-private; ingest's direct construction via `lakehouse.NewFileStore` is the intended call site.

rxbynerd:
## Summary

Introduces `stirrup-eval ingest` for the FileStore lakehouse. Each `--trace` file is detected per-format (streaming events vs. legacy single-blob) and ingested into `traces/` plus, for streaming files, `recordings/`. Interrupted streams are recorded with `Outcome=="interrupted"` by default; `--skip-partial` drops them.

This closes the loop on the demo narrative: `stirrup harness --trace ... && stirrup-eval ingest --trace ... --lakehouse ...` now populates both `traces/` and `recordings/` from real production runs, unblocking `mine-failures`, `replay`, and the `ReplayProvider`/`ReplayExecutor` paths end-to-end on local hardware.

Closes #271.
Part of #277.
Stacked on #283.

## Test plan

- [x] `just` passes
- [x] `just lint` passes
- [x] `TestIngestFile_StreamingFullRun` — happy path streaming → traces + recordings
- [x] `TestIngestFile_LegacySingleBlob` — backward-compat: legacy lines → traces only
- [x] `TestIngestFile_PartialStream` — interrupted run; both default ("interrupted" recording) and `--skip-partial`
- [x] `TestIngestFile_Idempotent` — three re-ingests of the same file produce one trace
- [x] `TestDetectFormat` — three-way classifier on streaming / legacy / empty

### DIFF
--- a/docs/eval.md
+++ b/docs/eval.md
@@ -424,6 +424,40 @@ rate, mean turns, p50/p95 duration) to stdout. Use this to seed an
 experiment baseline from real production data instead of static
 fixtures.
 
+### `ingest` — populate a lakehouse from JSONL traces
+
+```bash
+./stirrup-eval ingest \
+  --lakehouse var/lakehouse \
+  --trace tmp/sessions/run-1.jsonl \
+  --trace tmp/sessions/run-2.jsonl
+```
+
+Reads one or more JSONL trace files (produced by
+`stirrup harness --trace ...`) and persists them into a FileStore
+lakehouse. Two on-wire shapes are accepted transparently per file:
+
+- **Streaming event format (since #270)** — line-delimited events with
+  a `kind` discriminator. One file represents one run; ingest writes
+  both `traces/<runId>.json` and `recordings/<runId>.json`. Full
+  transcripts are preserved on the recording so replay and
+  mine-failures have something to chew on.
+- **Legacy single-blob format** — one `RunTrace` per line, no
+  discriminator. Each line ingests as one `traces/<id>.json` entry;
+  no recording is produced (the legacy shape has no transcript).
+
+Use `--trace -` to read from stdin (single file only). `--trace` is
+repeatable; mixed-format invocations are supported (per-file detection).
+
+A streaming trace that ended without a `run_finished` event (an
+interrupted run — SIGKILL, OOM) is ingested with
+`FinalOutcome.Outcome=="interrupted"` by default so it stays
+discoverable to mine-failures and replay. Pass `--skip-partial` to
+drop interrupted captures.
+
+Re-ingesting the same file is idempotent (last-write-wins, atomic
+rename via #267) so retries do not corrupt the lakehouse.
+
 ### `mine-failures` — turn production failures into tasks
 
 ```bash
--- a/eval/cmd/eval/ingest.go
+++ b/eval/cmd/eval/ingest.go
@@ -1,0 +1,343 @@
+package main
+
+// stirrup-eval ingest reads JSONL trace files produced by the
+// harness (`traceEmitter.type=jsonl`) and persists them into a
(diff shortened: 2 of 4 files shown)


rxbynerd:
this doesn't really make sense in the context of the prior PR (#283) as we now implement a brand-new write path outside of the lakehouse interface, but because more of this eval surface will require re-works later, I'll wrap that in as part of that work.

this is part of threading the line between "stirrup-eval being something useful for local development on stirrup" and "stirrup-eval as a first-class feature".

Issue #292, opened 2026-05-23
research: shared LLM client/provider library for stirrup + stirrup-eval


rxbynerd:
Spawned from review on #289. The diff-review judge had to re-implement an Anthropic `/v1/messages` HTTP client locally in `eval/judge/diffreview.go` because `eval/*` cannot import `harness/internal/provider/*` (a CLAUDE.md boundary the project enforces). The harness already has a well-tested provider stack — Anthropic, OpenAI Responses, Gemini, Bedrock, retry/replay/batch-poll wrappers, all behind a single `ProviderAdapter` interface — that the eval module would benefit from but cannot reach.

The duplication is small today (one model, one endpoint, ~80 lines) but compounds with every future LLM-driven judge (`spec-grader`, multi-model agreement, model-rated rubric scoring, etc.) and with every future provider added on the harness side. Each new wire-format surface added in `eval/` is one more place that has to learn snake_case quirks, retry/backoff posture, secret-redaction rules, and OTel `gen_ai.*` emission.

The review thread on #289 framed three plausible directions; this issue exists to research and decide between them before the next LLM-driven judge lands.

## Background

Current state:

- `harness/internal/provider/` is ~21 kLoC across four provider adapters (anthropic, openai, openai_responses, gemini, bedrock), plus retry, replay, and batchpoll wrappers, sitting behind the 14-line `ProviderAdapter` interface in `provider.go`. It depends on `harness/internal/credential` (secret resolution, GCP/Anthropic Workload Identity) and `harness/internal/observability` (OTel metrics + `gen_ai.*` attributes).
- `eval/judge/diffreview.go` re-implements a thin slice of the same surface against `api.anthropic.com/v1/messages` directly, with its own request/response types, its own 30-second timeout, and `ANTHROPIC_API_KEY`-only auth (no WIF support).
- `harness/internal/*` is intentionally private per [CLAUDE.md](../blob/main/CLAUDE.md). The public Go API surface is `harness/harnessapi/`, which does not currently expose the provider stack.
- The go.work layout is four sibling modules (`types`, `harness`, `eval`, `gen`). `types` is already the shared substrate for `StreamParams` / `StreamEvent` / `RunConfig`.

### A — Adopt an existing Go LLM client library

Survey the ecosystem (`anthropics/anthropic-sdk-go`, `tmc/langchaingo`, `openai/openai-go`, `sashabaranov/go-openai`, others) and adopt one that covers the providers the harness supports today.

[...]

Issue #130, opened 2026-05-09
github actions smoke test via anthropic workflow federation


rxbynerd:
## Background

Issue #16 called for an end-to-end smoke test against a real provider. Issue #117 delivered Anthropic Workload Identity Federation (WIF) support, which enables keyless authentication in GitHub Actions. This issue implements that smoke test.

## Goal

Create `.github/workflows/smoke-test.yml`: a manually dispatched (`workflow_dispatch`) workflow that builds the `stirrup` binary and runs `stirrup harness` against **Anthropic claude-haiku-4-5-20251001** using WIF — no static API key anywhere in the workflow.

Out of scope: evals, deeper testing, other providers. After the workflow passes, create follow-up issues for those.

## Preconditions (already done on the Anthropic Console side)

The federation rule `fdrl_01QJMFE86qFDdWzQXEekhYjY` is already configured to trust GitHub Actions OIDC tokens from `rxbynerd/stirrup`. No Console changes are needed.

## Deliverable

One new file: **`.github/workflows/smoke-test.yml`**

No other files need to change.

### Workflow design

| Property | Value |
|---|---|
| Trigger | `workflow_dispatch` only |
| Runner | `ubuntu-latest` |
| Permissions | `contents: read`, `id-token: write` |
| Model | `claude-haiku-4-5-20251001` |
| Max turns | `5` (keeps cost minimal) |
| Prompt | `Write the string 'smoke test passed' to a file called hello.txt` |
| Pass condition | Exit code 0 **and** `tail -1 trace.jsonl \| jq -e '.outcome == "success"'` |

### WIF credentials

The four federation identifiers are non-secret per Anthropic's docs and can be hardcoded directly in the workflow YAML (safe to commit):

| Field | Value |
|---|---|
| Federation rule ID | `fdrl_01QJMFE86qFDdWzQXEekhYjY` |
| Organization ID | `5848ccdc-495f-490b-9275-2caab9db344d` |
| Service account ID | `svac_01Tf7i7VwZm6WRAFjNAawPw9` |
| Workspace ID | _(omit — rule is bound to the default workspace only, so the field is optional and should be absent from the token exchange request)_ |

### CLI invocation

```sh
./stirrup harness \
  --anthropic-federation-rule-id fdrl_01QJMFE86qFDdWzQXEekhYjY \
  --anthropic-organization-id 5848ccdc-495f-490b-9275-2caab9db344d \
[...]

PR #294, opened 2026-05-23
eval: wire Anthropic WIF into eval-gate CI job

### LINKED ISSUE #293: eval: wire Anthropic WIF into eval-gate CI job
## Context

The eval-gate job in `ci.yml` currently authenticates to Anthropic via `ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}`. The project does not store a static API key as a repo secret — Anthropic access in CI is wired up via Workload Identity Federation (see #130 and `.github/workflows/smoke-anthropic.yml`), which is the correct posture for a public repository.

This is currently a latent gap: before `dogfood-seed.hcl` landed (PR #290), the `check-suites` step found no `.hcl` files and skipped the gate entirely. After #290 merges, the gate runs, `ANTHROPIC_API_KEY` is empty, all five harness invocations fail with an auth error, the `compare` step sees 0/5 vs the 5/5 baseline, and main's CI breaks.

**This must be resolved before or in the same merge as #290.**

It is independent of the #292 research deliverable (shared provider library) — WIF wiring is a CI concern, not a Go architecture concern.

### 1. `stirrup-eval run` — add WIF flag pass-through

The runner (`eval/runner/runner.go`) builds the harness `exec.Cmd` inline at lines 340–364. It currently passes `--config`, `--workspace`, `--trace`, `--prompt`, `--mode`, `--timeout`, but has no path for credential flags. The harness takes WIF credentials as CLI flags (see `smoke-anthropic.yml`):

```
--anthropic-federation-rule-id    <id>
--anthropic-organization-id       <id>
--anthropic-service-account-id    <id>
--anthropic-from-github-actions
```

`stirrup-eval run` needs corresponding flags forwarded verbatim to every harness invocation. The runner's `RunConfig` struct and `runTask` function are the places to extend.

### 2. `eval-gate` job in `ci.yml` — permissions + WIF flag arguments

The job needs:

```yaml
permissions:
  contents: read
  id-token: write        # needed to exchange OIDC token for Anthropic access token
```

And the `Run eval suite` step needs the four WIF flags passed to `stirrup-eval run`, analogous to the smoke test.

### 3. Add a credentials guard (short-term safety net)

To prevent future regressions where auth is not configured, add a pre-flight check analogous to the existing `check-suites` step — skip live eval runs if neither `ANTHROPIC_API_KEY` nor `--anthropic-from-github-actions` is configured, emitting a warning rather than a silent auth failure buried in per-task output.

## WIF identifiers

Non-secret, per Anthropic's WIF docs (already present in `smoke-anthropic.yml`):

| Flag | Value |
|---|---|
| `--anthropic-federation-rule-id` | `fdrl_01QJMFE86qFDdWzQXEekhYjY` |
| `--anthropic-organization-id` | `5848ccdc-495f-490b-9275-2caab9db344d` |
| `--anthropic-service-account-id` | `svac_01Tf7i7VwZm6WRAFjNAawPw9` |

## References

- `.github/workflows/smoke-anthropic.yml` — the existing WIF pattern this should mirror
- `eval/runner/runner.go` — `runTask` builds the harness invocation (lines 340–)
- `harness/cmd/stirrup/cmd/runconfigflags.go` — WIF flags on the harness side
- #130 — original WIF implementation
- #290 — `dogfood-seed.hcl` landing (the trigger for this issue going from latent to blocking)
- #292 — shared provider library research (independent; do not block on it)

rxbynerd:
Closes #293. Stacked on #290 (`feat/v01-13-first-mined-suite`).

## Why this PR exists

The eval-gate job in `ci.yml` currently authenticates to Anthropic via `ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}`, but the project does not store a static key as a repo secret — Anthropic access in CI is wired up via Workload Identity Federation (see #130 and `.github/workflows/smoke-anthropic.yml`). Once #290 lands `eval/suites/dogfood-seed.hcl`, the gate would discover the suite, find an empty key, fail all five tasks at the auth layer, and the `compare` step would report a 5/5 → 0/5 "regression" that is really an auth gap — breaking main on the merge.

This PR fixes that gap before #290 lands. It is independent of #292 (shared provider library research).

## Summary

- **`stirrup-eval run` (Go)**: four new flags (`--anthropic-federation-rule-id`, `--anthropic-organization-id`, `--anthropic-service-account-id`, `--anthropic-from-github-actions`) stored on `runner.AnthropicWIFConfig` and forwarded verbatim to every `stirrup harness` invocation built by `runTask`. The single arg-builder point covers both the legacy five-flag path and the merged-config path.
- **`eval-gate` job (`ci.yml`)**: gains `permissions.id-token: write` (the minimum scope the OIDC exchange needs) and passes the four `--anthropic-*` flags to the `Run eval suite` step. Identifier values are the same ones already in `smoke-anthropic.yml`.
- **Credentials guard**: a new `check-creds` step (analogous to `check-suites`) short-circuits the run/compare/upload steps with a `::warning::` annotation when neither `ANTHROPIC_API_KEY` nor WIF (`ACTIONS_ID_TOKEN_REQUEST_URL`) is configured. The guard checks the actual runtime environment variable GitHub Actions sets when `id-token: write` is in scope, so a future regression that strips the flags or drops the permission will produce a visible skip rather than a silent 0/N score.

## Test plan

- [x] `just build` — both binaries build clean
- [x] `just test` — full suite green including two new tests:
  - `TestAppendAnthropicWIFArgs` (table-driven, pins on-the-wire flag spellings)
  - `TestRunSuite_ForwardsAnthropicWIFFlags` (end-to-end fake-harness argv capture)
- [x] `just lint` — `golangci-lint v2` clean
- [x] `actionlint .github/workflows/ci.yml` — clean
- [ ] After merge: confirm the eval-gate job on main authenticates via WIF and the dogfood-seed suite runs end-to-end. (Cannot exercise on a PR branch — the job's `if:` gates on `refs/heads/main`.)

## Notes & deferred items

- The four WIF identifiers are non-secret per Anthropic's WIF docs. They are stored as plain values in `runner.AnthropicWIFConfig` (NOT in `types.RunConfig`, so the `Redact()` invariant is untouched). The actual OIDC exchange happens inside the harness subprocess using the GitHub Actions runner environment variables; no static key is in scope.
- The `diff-review` judge (`eval/judge/diffreview.go`) still reads `ANTHROPIC_API_KEY` from env. It is not exercised by `dogfood-seed.hcl` (which uses only `file-exists` / `file-contains` / `test-command` / `composite` judges), so this PR does not block on it. A follow-up to give the judge the same WIF treatment would be welcome but is out of scope here.
- The WIF identifier values are now duplicated between `smoke-anthropic.yml` and `ci.yml`. A future DRY pass (reusable workflow or composite action) is worth considering if a third caller appears.
- The `actions/checkout@v6` and `actions/setup-go@v6` references in `eval-gate` are floating-tag (not SHA-pinned), which is inconsistent with the `publish-container` job convention. That pre-existing gap is out of scope here; tracked separately.

Refs #293, #290, #130.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,17 @@ jobs:
     runs-on: ubuntu-latest
     needs: verify
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    # id-token: write is required for Anthropic Workload Identity
+    # Federation: actions/checkout's runner exposes
+    # ACTIONS_ID_TOKEN_REQUEST_URL / ACTIONS_ID_TOKEN_REQUEST_TOKEN
+    # only when this permission is granted. The harness then exchanges
+    # the OIDC token for an Anthropic access token using the four
+    # `--anthropic-*` flags forwarded below. No static API key is in
+    # scope. See .github/workflows/smoke-anthropic.yml for the
+    # established WIF pattern and #130 for the original rollout.
     permissions:
       contents: read
+      id-token: write
     steps:
       - name: Check out repository
         uses: actions/checkout@v6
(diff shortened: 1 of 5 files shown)


rxbynerd:
This is clumsy, and justifies #292.

PR #316, opened 2026-05-24
provider/quirks: wire openai-compatible quirks + Z.ai compat + Responses plumbing (Wave 2 Step 2+4)

rxbynerd:
Second integration PR for #221 / Wave 2 of the tool-use debacle workstream. Builds on PR #315 (scaffold). Wires the quirks layer into the openai-compatible adapter, adds the first builtin rules, introduces the Z.ai compat profile, and plumbs the Responses adapter. **Behaviour change**: reasoning-class models (o-series, GPT-5 non-chat) now omit sampling params automatically; `CompatProfile = "zai-glm"` enables Z.ai's `max_tokens` + `tool_stream` extension.

Per the approved Wave 2 design at `stirrup-coord/scratch/wave-2-design.md`.

### openai-compatible adapter (`provider/openai.go`)
- `Stream()` calls `Registry.ResolveWithRules(providerType, params.Model)` at the top; resolved `ProviderQuirks` and the matching rule list flow into `buildOpenAIRequest`.
- `openaiRequest` refactored to custom `MarshalJSON`/`UnmarshalJSON` — dynamic field inclusion/exclusion (`OmitSamplingParams`), token-field switching (`TokenFieldMaxCompletionTokens` ↔ `TokenFieldMaxTokens`), and `ExtraBodyFields` merge can't be expressed with static struct tags.
- `MarshalJSON` errors on `ExtraBodyFields` key collision with canonical fields; `UnmarshalJSON` errors if both `max_completion_tokens` and `max_tokens` keys appear in the same body (instead of silently overwriting).
- Per-`Stream` debug log lists applied rule descriptions; warn log on suppressed temperature references the rule. Suppressed value not logged (sidechannel concern).

### Builtin rules (`provider/quirks/builtin.go`)
- `o[1-9]*` for openai-compatible → `applyOpenAIReasoningClass` (covers bare `o3`, `o4`, dash-suffix `o1-mini`, and forward-compat two-digit `o10-mini`).
- `gpt-5*` → `applyOpenAIReasoningClass`.
- `gpt-5-chat*` → carve-out: clears `OmitSamplingParams` so chat models keep `temperature`/`top_p`/etc. Longest-glob-wins composition (D10) puts the carve-out last.

### Z.ai compat profile (`provider/compat/zai/`)
- `CompatRule()` exported: matches Z.ai GLM models; sets `TokenField = TokenFieldMaxTokens` and `ExtraBodyFields[\"tool_stream\"] = true`.
[...]

Issue #231, opened 2026-05-19
tools: return structured tool results with text fallback


rxbynerd:
## Type

Feature request

## Motivation

Most Stirrup tool results are currently returned as one text string. That works for broad provider compatibility, but it makes tool outputs harder for models and downstream systems to parse reliably. Command results, search hits, and file excerpts are all better represented with stable fields such as path, line, text, stdout, stderr, and exit code.

Structured results should improve model reliability while preserving a text fallback for providers that only accept string tool results.

## Current state

- `types.ToolResult` has `Content string` and `IsError`.
- `run_command` concatenates stdout, stderr, and exit code into a string.
- `search_files` returns raw grep/find output.
- `read_file` returns raw file content.
- MCP `tools/call` handling preserves text content only and drops non-text content, structured content, resource links, output schemas, and annotations.

## Goals

1. Add a structured result representation alongside the existing text content.
2. Preserve text fallback for all providers.
3. Return structured command results, including `stdout`, `stderr`, `exit_code`, and timeout metadata.
4. Return structured search matches, including path, line, column if available, and match text.
5. Return structured file excerpts, including path, start line, end line, truncation state, and text lines.
6. Preserve richer MCP result content where supported.

## Non-goals

- Requiring all providers to accept non-string tool results immediately.
- Changing the executor API more than needed.
- Building a UI for structured results.

## Design considerations

- Keep `Content string` for backwards compatibility and provider fallback.
- Consider adding `Structured json.RawMessage` or a typed result envelope to `types.ToolResult`.
- Provider adapters should serialize structured results only when the provider supports the shape; otherwise they should send the text fallback.
- Trace output should include structured data when safe and scrubbed.
- MCP support should track the current MCP tool result spec, including `structuredContent`, resource links, embedded resources, and non-text content.

## Acceptance criteria

- `ToolResult` can carry structured data without breaking existing text-only flows.
- Built-in command/search/read results populate structured data and text fallback consistently.
- Provider adapters preserve compatibility with current string result behavior.
- MCP bridge preserves at least text plus structured content where available; unsupported content types are represented or explicitly marked rather than silently discarded.
- Tests cover provider serialization, built-in tool structured output, and MCP structured output handling.
- Docs describe the result envelope and provider fallback behavior.
- `go test ./harness/... ./types/...` passes.

## References

- `types/messages.go` — `ToolResult`
- `harness/internal/tool/builtins/shell.go`
- `harness/internal/tool/builtins/search.go`
[...]

Issue #228, opened 2026-05-19
provider: add schema linting and strict-mode normalization for tool inputs


rxbynerd:
## Type

Feature request

## Motivation

Strict tool-calling modes can improve argument accuracy, but Stirrup's current schemas are not uniformly strict-mode ready. Some providers require specific schema shapes, and some optional-field patterns in current tools would be rejected or silently changed by strict modes.

Before Stirrup enables strict mode broadly, it needs provider-specific schema linting and normalization so failures happen clearly and early.

## Current state

- Tool inputs are JSON Schemas stored as raw JSON.
- Central runtime validation uses JSON Schema Draft 2020-12.
- Provider adapters translate schemas directly or through provider-specific conversion.
- Gemini conversion already rewrites/drops unsupported schema features to fit Vertex AI function declarations.
- Current built-in schemas include optional fields not listed in `required`, such as `search_files.path`, `run_command.timeout`, and most `edit_file` mode fields.
- OpenAI strict mode requires all properties to be listed as required, with nullable types used to represent optional semantics.

## Goals

1. Add provider-specific schema linting before tool definitions are sent to a provider.
2. Detect unsupported JSON Schema keywords per provider/profile.
3. Add a strict-mode normalizer for providers that require all fields to be required with nullable optional fields.
4. Make OpenAI Responses `strict` behavior explicit so optional-field behavior is not accidental.
5. Emit clear errors that identify the tool name, schema path, provider, and incompatible keyword/shape.

## Non-goals

- Redesigning every built-in schema in this issue.
- Validating MCP server schemas at trust level higher than remote/untrusted; MCP schemas should still be treated cautiously.
- Implementing provider capability profiles; this issue consumes those once available.

## Design considerations

- Keep the canonical internal schema separate from provider-normalized schemas.
- Do not mutate the original `json.RawMessage` in the registry.
- Cache normalized schemas per provider/profile to avoid repeated work every turn.
- Fail closed when strict mode is requested and normalization is impossible.
- Add regression tests for optional fields, additionalProperties, unsupported unions, `$ref`, `allOf`, and Gemini-specific conversions.

## Acceptance criteria

- Provider schema linting runs before request serialization for tool-enabled calls.
- Strict-mode normalization can convert safe optional fields to provider-compatible nullable-required fields.
- Unsafe/incompatible schemas fail with precise errors.
[...]

Issue #227, opened 2026-05-19
tools: enrich built-in tool descriptions with examples and usage guidance


rxbynerd:
## Type

Feature request

## Motivation

Provider docs for tool calling consistently emphasize clear tool names, detailed descriptions, parameter descriptions, and examples. Stirrup's built-in tool descriptions are currently concise, but they do not give models enough guidance about when to use each tool, what the result looks like, or common invalid argument patterns.

This matters for multi-provider operation because not every model has the same priors around coding harness tools.

## Current state

- Built-in tools define short descriptions in `harness/internal/tool/builtins/*` and `harness/internal/edit/*`.
- `types.ToolDefinition` does not currently support provider-native `input_examples`, so examples would need to live in descriptions or wait on the provider-facing tool contract work.
- Prompt templates do not appear to contain a detailed tool-use inventory for the default built-in tools.

## Goals

1. Add high-quality, concise descriptions for each built-in tool.
2. Include guidance for:
   - when to use the tool
   - when not to use the tool
   - required arguments
   - what the result looks like
   - common mistakes that should be avoided
3. Add one concise example input per tool, either in the description or through the richer tool contract once available.
4. Ensure descriptions remain short enough not to waste context unnecessarily.
5. Add tests or snapshots so descriptions do not accidentally regress to ambiguous text.

## Non-goals

- Changing tool behavior.
- Adding provider-specific aliases.
- Implementing strict mode. This issue only improves the model's understanding of existing tools.

## Design considerations

- Once #222 exists, map examples into Anthropic `input_examples` and any equivalent provider-native surfaces.
- Until then, a disciplined description format is acceptable.
- Avoid long prose; models benefit from precise, compact instructions.
- Include output-shape promises only if the tool actually guarantees them.
- Keep security-sensitive guidance out of descriptions if it would help prompt injection bypass policy.

## Acceptance criteria

- Every built-in tool has a reviewed description with usage guidance and an example input.
- Descriptions for read/search/edit tools explain common workflows clearly.
- Descriptions for sensitive tools such as `run_command`, `web_fetch`, and `spawn_agent` describe their effect and approval/safety posture accurately.
- Provider adapters preserve the improved descriptions without truncation or malformed JSON.
- Tests or snapshots detect accidental removal of examples/guidance.
- Docs mention the conventions for future tool descriptions.

## References

- `harness/internal/tool/builtins/filesystem.go`
- `harness/internal/tool/builtins/search.go`
[...]

PR #319, opened 2026-05-24
tool/builtins: redesign read_file, split search_files, require edit_file operation enum (Wave 3 Step A)

### LINKED ISSUE #225: tools: redesign built-in coding tool schemas for lower-friction agent use
## Type

Feature request

## Motivation

The built-in coding tools work, but they make the model infer too much. Multiple providers struggle with the current surface because common coding-agent actions are not first-class enough:

- `read_file` can only read a whole file and returns no line numbers.
- `search_files` overloads content search and file-name search behind a required `type` field.
- `search_files` shells out to `grep`/`find` and has no result limits or include/exclude controls.
- `edit_file` infers mode from optional fields instead of requiring an explicit operation.

The goal is to make the default toolset easier for a range of models to use correctly, not only models that have been heavily tuned on one tool vocabulary.

## Current state

- `harness/internal/tool/builtins/filesystem.go` defines `read_file`, `write_file`, and `list_directory` schemas.
- `read_file` input is only `{ "path": string }`.
- `list_directory` input is only `{ "path": string }` and returns a newline list.
- `harness/internal/tool/builtins/search.go` defines `search_files` with `pattern`, optional `path`, and required `type: grep|glob`.
- `harness/internal/edit/multi.go` defines `edit_file` with only `path` required and optional `diff`, `old_string`, `new_string`, and `content` fields.

## Goals

1. Add bounded, line-numbered file reading.
   - Add `start_line` and `limit` (or equivalent) to `read_file`.
   - Return stable line-numbered excerpts.
   - Preserve the existing whole-file behavior where safe and within file-size limits.
2. Improve directory discovery.
   - Consider `recursive`, `max_depth`, or `max_entries` for `list_directory`.
   - Keep output stable and bounded.
3. Redesign search for coding workflows.
   - Prefer `rg` when available, with fallback where necessary.
   - Add include/exclude filters, max result count, and optional context lines.
   - Either split content search and file-name search into separate tools, or make the mode field explicit and less error-prone.
4. Make edit mode explicit.
   - Add an `operation` enum to `edit_file`, or expose separate strict-friendly edit tools.
   - Validate incompatible combinations before attempting edits.
5. Maintain backwards compatibility or provide clear migration aliases for existing configs.

## Non-goals

- Changing executor sandbox semantics.
- Replacing the edit strategies themselves unless required by the schema redesign.
- Adding provider-specific tool names in this issue; provider-friendly aliases are tracked separately.

## Design considerations

- Schema changes should be compatible with future strict-mode support (#222 and the schema-linting issue).
- Tool results should be stable enough for models to parse: file path, line numbers, and truncation markers should not vary ad hoc.
- Search output must be bounded so a broad grep cannot flood context.
- `rg` is preferred for developer ergonomics, but container/API executors may not have the same capabilities; keep fallbacks explicit.
- If backwards compatibility requires accepting old inputs, normalize old inputs into the new operation internally and warn in traces/logs.

## Acceptance criteria

- `read_file` supports bounded line-range reads and returns line-numbered output.
- Search supports bounded results plus include/exclude controls and prefers `rg` when available.
- `edit_file` has an explicit operation or is replaced by separate strict-friendly edit tools.
- Invalid tool argument combinations fail schema validation or a clear preflight validation before execution.
- Existing read/search/edit tests are updated, and new tests cover model-common mistakes.
- Documentation and tool descriptions explain the new schema and output shape.
- `go test ./harness/... ./types/...` passes.

## References

- `harness/internal/tool/builtins/filesystem.go`
- `harness/internal/tool/builtins/search.go`
- `harness/internal/edit/multi.go`
- `harness/internal/core/factory.go::buildToolRegistry`
- #222 — provider-facing tool contract improvements

rxbynerd:
Closes #225. First PR of Wave 3 of the tool-use debacle workstream. **Stacked on [PR #318](https://github.com/rxbynerd/stirrup/pull/318)** (Wave 2 final). The full stack: main → #315 (merged) → #316 → #317 → #318 → this PR. Subsequent Wave 3 steps (#228 strict-mode lint, #227 enriched descriptions) will stack on this.

## Breaking changes (clean break per workstream §6.4 — pre-1.0)

- **`search_files` is removed.** No backwards-compatibility alias. Operators must update RunConfigs to use `grep_files` (regex content search) and/or `find_files` (glob filename search).
- Models that call `search_files` get an explicit migration error: `tool not found: search_files; use grep_files (regex content search) or find_files (glob filename search)`.
- **`edit_file.operation` is now required.** Old configs that relied on operation inference from optional fields must add the explicit `operation` field (`replace` / `delete` / `rewrite` / `patch`).

### `read_file` — line ranges + numbered output
- New params: `start_line`, `limit` (or equivalent). Default returns the whole file.
- Output prefixed with line numbers.
- `start_line` past EOF returns empty content + a clear notice (NOT an error) — forgiving model usage.

### `search_files` split into two strictly-typed tools
- **`grep_files`** — regex content search. Prefers `rg` (ripgrep) when available + `CanExec=true`; falls back to native Go `regexp` + `filepath.WalkDir`. Bounded by `max_results` (default 100, max 1000). Outputs `path:line:match` lines. `include`/`exclude` glob filters.
- **`find_files`** — glob filename search. Pure Go (`filepath.WalkDir` + `filepath.Match`). Bounded by `max_results`. Schema description explicitly states basename-only matching (use `include` for path-substring filtering). `include`/`exclude` glob filters.
- Both tools require only `caps.CanRead` (not `caps.CanExec`) — sandboxed read-only executors retain access. `grep_files` still uses the rg fast-path when `CanExec=true`.

### `edit_file` — explicit operation enum
- `operation` is required, one of `replace`/`delete`/`rewrite`/`patch`.
- Per-operation field validation enforces required-field contracts (e.g. `replace` requires `old_string` + `new_string`; `delete` requires `old_string` only).
- Chose 4 values not 5: `insert` would require a new edit strategy (out of scope); use `replace` with positional context or `patch` with a udiff instead.

### `list_directory` — bounded recursion
- New params: `recursive` bool, `max_depth` (default 3, max 10), `max_entries` (default 1000, max 10000). BFS traversal; depth-bounded regardless of symlink chains.

### Read-only-mode allow-list
- `grep_files`, `find_files` added to `DefaultReadOnlyBuiltInTools`. `planning`/`review`/`research`/`toil` modes can use them. `mutatingTools` unchanged.

### Example RunConfigs + proto + safety/security docs
- 14 example RunConfigs updated to reference `grep_files`/`find_files` (would otherwise fail `ValidateRunConfig`).
- Proto field docstring updated.
- `docs/safety-rings.md` (4 inline example sites) and `docs/security.md` updated to drop stale `search_files` references; security guarantee descriptions cover both new tools.
- `docs/architecture.md` + `docs/configuration.md` minor permission-table/tool-list refresh.

## Pre-merge remediation

Reviewers caught 9 blockers + 4 polish items; all resolved.

**Two critical security/correctness fixes flagged separately for reviewer attention:**

1. **CWE-59 file exfiltration via symlinks (HIGH/blocking, security reviewer).** `grepNative` was calling `os.ReadFile` on raw `filepath.WalkDir` paths, which followed symlinks unconditionally. A symlink inside the workspace pointing to `/etc/shadow` (or any external file) would have its content returned to the model when matching the grep pattern. Fixed with a one-line `d.Type() & fs.ModeSymlink != 0` skip in the WalkDir callback. `TestGrepNative_SkipsSymlinks` verified to FAIL on the pre-fix code (the symlinked content leaked) and PASS post-fix.

2. **CWE-185 filter fail-open in include/exclude globs (HIGH/blocking, code+security+api).** The hand-rolled `doubleStarMatch` glob→regex translator only escaped `.`; other regex metacharacters (`[`, `]`, `(`, `)`, `+`, `{`, `}`, `|`, `\`, `^`, `$`, `?`) passed through verbatim, and iteration was by byte not rune so multi-byte UTF-8 paths produced invalid regex fragments. Filters silently failed open (included files that should have been excluded). Fixed by rune iteration + `regexp.QuoteMeta` on default-branch characters. `TestDoubleStarMatch_EscapesMetacharacters` had 8/14 subcases failing pre-fix; all pass post-fix.

**Other blockers resolved:**

3. `find_files` schema description rewritten to state basename-only semantics explicitly; misleading `**/handler_*.ts` example replaced with `handler_*.ts`.
4. `grep_files`/`find_files` capability gate moved from `CanExec` to `CanRead`. Test seam in `integration_test.go` flipped from assert-absent to assert-present.
5. `grepViaRipgrep` transport errors (`err != nil`) now fall back to the native walker, matching the existing exit-code-≥2 fallback. Context cancellation still propagates.
6. `validateOperationFields` `replace`/`delete` missing-`old_string` branches now have direct tests.
7. Migration hint test pins the descriptive suffixes (`regex content search` / `glob filename search`) so a silent reword fails the test.
8. `ripgrepDetector.detect` once-cache behaviour pinned with a probe-counter test.

**Non-blocking polish applied in this PR:**

- `TestRegisterBuiltins` asserts `search_files` is ABSENT.
- `renamedToolHint` wording pinned by direct unit test.
- `grepViaRipgrep` include/exclude `--glob` arg construction asserted.
- `grepNative` / `findNative` `fs.ErrPermission` skip paths covered.

## Verification

- `just` (build + full test suite) exits 0.
- `just lint` (golangci-lint v2) reports `0 issues`.
- Coverage on new tool implementations: `ReadFileTool` 95.5%, `formatReadFile` 100%, `ListDirectoryTool` 87.9%, `listDirectoryRecursive` 92%, `GrepFilesTool` 93.1%, `FindFilesTool` 95.5%, `MultiStrategy` (edit) 89.3% — all above the 85% bar.
- Critical-finding fixes verified by stash-and-retest: pre-fix code fails the regression test, post-fix passes.

## Deferred (Step B / Step C / Wave 4 — NOT in this PR)

- Strict-mode normalization (`InputSchema` lint per provider) — Step B / #228 (next stacked PR). The seam is identified: `validateOperationFields` runtime check can be lifted to provider-side JSON Schema `allOf`/`oneOf`; enum declarations and bound declarations are in place.
- Enriched descriptions / `input_examples` — Step C / #227 (final Wave 3 PR).
- Structured tool results (`Structured json.RawMessage` envelope on `ToolResult`) — Wave 4 / #231.
- `insert` operation on `edit_file` — out of scope per ambiguity resolution.
- Path-substring `find_files` matching — basename-only is the chosen ergonomic; use `include` for path filtering.

Reviewed by code, security, spec-compliance, test-coverage, and api-design reviewers; remediation brief in workstream scratch. Verdict: NEEDS REMEDIATION (applied).


harness/internal/tool/builtins/search.go
rxbynerd, line 1: filename should be changed to grep

rxbynerd: nvm contains both grep tool and find tool, so in future we split into two files

rxbynerd (commented):
intersection of ripgrep plus permissionpolicy should be confirmed before release: edge case may exist where permissionpolicy denies access to a file but `find_files` will still return it

PR #321, opened 2026-05-24
tool/builtins: enrich descriptions with when-to-use guidance and worked examples (Wave 3 Step C)

rxbynerd:
Closes #227. **Final Wave 3 PR.** Stacked on [PR #320](https://github.com/rxbynerd/stirrup/pull/320) (Wave 3 Step B). The full stack: main → #315 (merged) → Wave 2 [#316, #317, #318] → Wave 3 [#319, #320, this PR].

When this stack merges, the tool-use debacle workstream's Waves 1–3 are complete; remaining waves cover loop recovery (#230/#231), eval suite (#233), aliases (#234), and lazy loading (#235).

---

## Wave 3 summary (across #319, #320, this PR)

Wave 3 reshapes the built-in tool surface so models can use it correctly without trial and error and providers can lint/enforce it server-side:

- **Step A (#319)** — schema redesign. `read_file` gets line ranges + numbered output. `search_files` splits into `grep_files` (regex content) + `find_files` (glob filename, basename-only) with no BC alias; legacy callers see an explicit migration error. `edit_file` requires an explicit `operation` enum (`replace`/`delete`/`rewrite`/`patch`) with per-op required-field validation. `list_directory` gets bounded recursion. Critical security fix during remediation: `grep_files` native walker no longer follows workspace-escaping symlinks (CWE-59), and glob include/exclude filters no longer fail open via unescaped regex metacharacters (CWE-185). 5 example RunConfigs + safety/security docs updated for the rename.

- **Step B (#320)** — provider-side schema lint + OpenAI strict-mode normalisation. New `quirks/strictschema.go` recursively rewrites tool schemas to OpenAI's strict-mode shape (all-required, nullable optionals, no additionalProperties); fails closed on unsupported keywords; cached per-adapter under singleflight semantics. New `gemini_schema_lint.go` rejects schemas containing keywords the resolved Gemini profile doesn't accept. Quirks rules opt in `gpt-4o-mini*` / `gpt-4.1*` / `gpt-5*` (with the `gpt-5-chat*` carve-out preserving strict mode) and Gemini-3 lint. Error messages identify tool + field path, never leak `description` or `enum` content.

[...]

Issue #313, opened 2026-05-24
observability: reserve no_tool_when_required category for Wave 4 #230 integration


rxbynerd:
Pre-stabilisation cleanup of the `ToolFailureCategory` enum added by #304, ahead of Wave 4's tool-choice escalation work (#230).

## Issue

`harness/internal/observability/toolfailure.go` declares a closed 19-value enum guarded by `IsValid()`. #230 will add a tool-choice escalation mechanism for runs where the model skips required workspace context — the natural failure category name is `no_tool_when_required`.

Reserving the constant *now* (before the enum is considered stable) means:
- Wave 4 only wires the emission site, not the enum addition + emission together.
- `TestToolFailureCategory_EnumIsValid` and downstream dashboards have prior notice.
- The enum's wire format does not gain a value mid-release-cycle for an existing-category dashboard.

## Fix

Add to `harness/internal/observability/toolfailure.go`:

```go
// ToolFailureNoToolWhenRequired — model returned without calling any
// tool when the harness required tool use (Wave 4 / #230). No emission
// site exists in this binary; reserved for forward compatibility.
ToolFailureNoToolWhenRequired ToolFailureCategory = \"no_tool_when_required\"
```

Add the matching entry to `allToolFailureCategories` so `IsValid()` accepts it. `TestToolFailureCategory_EnumIsValid` will catch any omission.

No emission site is needed in this PR — that wires up in #230.

## Why deferred

Forward-compatibility-only addition; api-design-reviewer flagged that adding it before the enum stabilises avoids a mid-cycle wire-format change. Per the wave-1-issue-229 synthesis deferred-items list.

Followed up from #304. Coordinates with #230.

Issue #230, opened 2026-05-19
core: add tool-choice escalation when models skip required workspace context


rxbynerd:
## Type

Feature request

## Motivation

A common failure mode for autonomous coding agents is giving a final answer without inspecting the workspace, even when the task clearly requires tool use. Stirrup currently relies on prompt instructions and provider defaults. It does not have a recovery path that escalates from normal `auto` tool choice to provider-native required/validated tool choice, or to a stronger retry prompt, when a no-tool turn is suspicious.

This issue adds a bounded recovery mechanism for missed tool use.

## Current state

- The core loop dispatches tools only when provider output has a tool-use stop reason / tool calls.
- If the provider returns `end_turn` with text, the loop accepts it as a final answer.
- Provider adapters do not expose tool-choice controls through `StreamParams` today.
- The richer tool contract / provider capability work is tracked separately and should provide the provider-native controls consumed here.

## Goals

1. Detect likely no-tool failures for tasks that require workspace context.
2. Retry once, or a bounded number of times, with stronger tool-use instructions or provider-native required/validated tool mode where supported.
3. Make the behavior configurable and safe for modes where no tool is required.
4. Record metrics/traces for tool-choice escalation.
5. Avoid infinite loops or repeated costly retries.

## Non-goals

- Forcing tool use on every turn.
- Replacing prompts with a planner.
- Solving provider tool-choice serialization; this consumes #222/provider profile work when available.

## Design considerations

- Start conservative. Example trigger: coding/review/execution task, tools are available, no prior tool calls happened, and the assistant attempts a final answer.
- Allow mode-specific policy: planning/review may require at least one read/search; research may allow web_fetch; execution may require read before edit.
- When provider-native controls are unavailable, use a retry prompt that states the missing requirement and asks for an appropriate tool call.
- Cap escalation to one retry by default.
- Include a clear trace event so operators can audit why the retry happened.
- Be careful not to override legitimate no-tool answers for purely conceptual questions.

## Acceptance criteria

- The loop can detect and recover from a first-turn no-tool answer on a workspace-dependent task.
- Provider-native required/validated tool-choice is used when supported by the selected profile.
- A prompt-only fallback is used when provider-native control is unsupported.
- Escalation is bounded and cannot cause an unbounded retry loop.
- Metrics/traces record escalation attempts and outcomes.
- Tests cover: no-tool final answer -> retry; legitimate no-tool answer -> no retry; provider-native path; prompt fallback path; max retry cap.
[...]

PR #322, opened 2026-05-24
core: tool-choice plumbing — StreamParams control + cross-provider quirks capability (Wave 4 A1)

rxbynerd:
## Summary

First chunk of **Wave 4 / #230** (tool-choice escalation). This PR is the **plumbing layer only** — it makes provider-native tool-choice *expressible and serializable*, gated on a provider capability. The loop-side escalation logic (no-tool-turn detection, bounded retry, mode-aware policy) is the next chunk (A2) and is deliberately not here.

Stacked on the Wave 3 stack: `#321` ← **A1**. Auto-retargets to `main` as lower PRs merge.

## What lands

- **`types.StreamParams`** gains a closed `ToolChoiceMode` enum (`auto`/`required`/`none`/`tool`) + `ToolChoiceName`. Zero value = `auto` = no wire field, so every existing caller is byte-identical.
- **`ProviderQuirks.ToolChoice`** — a new top-level cross-provider `ToolChoiceCapability` (per-mode bools), not buried under an adapter sub-struct, because tool-choice spans Anthropic/OpenAI/Gemini and the BehaviourFlags ownership rule forbids cross-family reads. Three builtin `*`-glob rules (anthropic / openai-compatible / gemini).
- **Serialization** in all three adapters, gated on the resolved capability (emit nothing when unsupported — graceful no-op; the prompt fallback is A2's job):
  - Anthropic `tool_choice` object (`any`/`tool`),
  - OpenAI `tool_choice` (`required`/`none`/typed function object),
  - Gemini `functionCallingConfig.mode` (`ANY`/`NONE`) + `allowedFunctionNames`.

## Review hardening (post-review remediation)

- Typed `openAINamedToolChoice` replaces a `map[string]any` (deterministic JSON key order + project no-`any` rule); pinned by a 20× marshal-determinism test.
- `ValidateToolChoiceName` (`^[a-zA-Z0-9_-]{1,64}$`) at the wire boundary — invalid names **degrade** to the auto/no-field result with a `slog.Warn` that logs grammar + **length only**, never the raw name (avoids log-injection once A2 feeds model-influenced names through).
- Partial-capability tests bring all three `*ToolChoiceFromParams` projections to 100% function coverage.

## Test plan

- [x] `just` (build + test) green
- [x] `just lint` (golangci-lint v2) — 0 issues
- [x] Zero-value (auto) emits no `tool_choice` on any adapter — byte-identical to pre-#230
- [x] Per-adapter, per-mode serialization asserted against marshalled request bodies (no live provider calls)
- [x] Capability-gating + partial-capability omission covered
- [ ] Consumed end-to-end by A2 (next chunk)

Closes part of #230 (plumbing). Loop escalation + `no_tool_when_required` (incl. #313) follow in A2.

### DIFF
--- a/harness/internal/provider/quirks/builtin.go
+++ b/harness/internal/provider/quirks/builtin.go
@@ -238,6 +238,73 @@ func BuiltinRules() []Rule {
 				)
 			},
 		},
+		// --- Tool-choice capability rules (#230, Wave 4 A1) ---
+		//
+		// tool_choice is a cross-provider capability, so the resolved
+		// flag lives on the top-level ProviderQuirks.ToolChoice field
+		// rather than under a provider sub-struct. Each first-party
+		// provider declares a base "*" rule advertising the modes its
+		// API supports; adapters gate serialisation on the resolved
+		// flag and emit nothing when Supported is false. The escalation
+		// chunk (A2) consumes the StreamParams.ToolChoice control these
+		// rules make safe to serialise.
+		{
+			ProviderType: "anthropic",
+			ModelMatch:   "*",
+			Description:  "Anthropic: native tool_choice (auto/any/tool); no native none",
+			LastVerified: Date("2026-05-24"),
+			Apply: func(q *ProviderQuirks) {
+				// Anthropic's Messages API accepts a tool_choice object
+				// with type "auto", "any", or "tool". There is no native
+				// "none" — a no-tools turn is expressed by omitting the
+				// tools array — so None stays false and the adapter
+				// handles ToolChoiceNone structurally.
+				q.ToolChoice = ToolChoiceCapability{
+					Supported: true,
+					Auto:      true,
+					Required:  true,
+					None:      false,
+					NamedTool: true,
+				}
+			},
+		},
+		{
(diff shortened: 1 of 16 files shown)


harness/internal/provider/quirks/builtin.go
```diff
@@ -238,6 +238,73 @@ func BuiltinRules() []Rule {
...
+		// tool_choice is a cross-provider capability, so the resolved
+		// flag lives on the top-level ProviderQuirks.ToolChoice field
+		// rather than under a provider sub-struct. Each first-party
+		// provider declares a base "*" rule advertising the modes its
+		// API supports; adapters gate serialisation on the resolved
+		// flag and emit nothing when Supported is false. The escalation
+		// chunk (A2) consumes the StreamParams.ToolChoice control these
+		// rules make safe to serialise.
+		{
+			ProviderType: "anthropic",
```
rxbynerd, line 252: basing this on the provider will likely create a divergence between anthropic-direct & anthropic-bedrock calls, as I doubt this would trip for bedrock.

PR #328, opened 2026-05-25
tools: complete the model-facing tool contract (#222)

### LINKED ISSUE #222: tools: extend model-facing tool definitions with provider reliability controls
## Type

Feature request

## Motivation

Stirrup's internal tool execution model is richer than the tool definition exposed to providers. The model-facing `types.ToolDefinition` contains only `Name`, `Description`, and `InputSchema`, so adapters cannot express modern provider controls such as strict schema mode, tool choice, parallel tool-call policy, input examples, annotations, output schemas, or provider-specific tool fields.

That makes Stirrup a lowest-common-denominator tool caller. When a model emits malformed arguments, skips a required tool, or calls tools too aggressively, the harness can only recover after the fact instead of using provider-native controls up front.

## Current state

- `types/messages.go` defines `ToolDefinition` with only three fields.
- `harness/internal/tool/tool.go::Definition()` drops local execution metadata and has no place for provider-facing metadata.
- `harness/internal/provider/openai.go::translateTools` copies only name/description/parameters.
- `harness/internal/provider/openai_responses.go::translateToolsResponses` copies only name/description/parameters.
- `harness/internal/provider/anthropic.go` passes the minimal tool definitions directly.
- `harness/internal/provider/gemini_request.go` converts only name/description/parameters to function declarations.

## Goals

1. Extend Stirrup's model-facing tool contract or add an adapter-side capability layer that can represent:
   - strict schema mode
   - tool choice (`auto`, `none`, `required`, specific tool where supported)
   - parallel tool-call policy
   - input examples
   - tool annotations
   - output schemas / structured result metadata
   - provider-specific tool properties where necessary
2. Preserve the current simple built-in tool implementation API where possible.
3. Make provider adapters explicitly map supported fields and intentionally ignore unsupported fields.
4. Make OpenAI Responses `strict` behavior explicit so optional-field behavior is intentional.

## Non-goals

- Redesigning every built-in tool schema. That is tracked separately.
- Enabling strict mode for all current tools immediately. Strict readiness needs schema linting/normalization first.
- Replacing MCP's protocol schema. MCP bridge improvements can consume this contract later.

## Design considerations

- Avoid stuffing every provider-specific option into `types.ToolDefinition` if a smaller generic `ToolBehavior` / `ToolPresentation` struct would keep the model clean.
- The local `tool.Tool` execution flags (`WorkspaceMutating`, `RequiresApproval`) are not necessarily safe to expose directly to providers, but provider-facing annotations may be derived from them.
- OpenAI strict mode has schema-shape requirements; Anthropic strict mode has a supported subset; Gemini uses function-calling modes. The contract should support per-provider support checks.
- This issue should be implemented with request-shape unit tests rather than relying on live provider calls.

## Acceptance criteria

- Tool definitions can carry strict/tool-choice/parallel-call metadata through the core registry to adapters.
- OpenAI Chat and OpenAI Responses adapters serialize supported reliability fields correctly.
- Anthropic adapter serializes supported fields such as `strict` and examples where applicable.
- Gemini adapter can select an appropriate function-calling mode based on the tool policy.
- Unsupported provider/tool metadata is ignored or rejected deliberately with a test-covered error path.
- Existing tool registry and provider tests continue to pass.
- Docs explain which tool reliability controls are available per provider.

## References

- `types/messages.go` — `ToolDefinition`
- `harness/internal/tool/tool.go` — local tool metadata and `Definition()`
- `harness/internal/provider/openai.go::translateTools`
- `harness/internal/provider/openai_responses.go::translateToolsResponses`
- `harness/internal/provider/anthropic.go` — Anthropic request shape
- `harness/internal/provider/gemini_request.go` — Gemini function declarations
- OpenAI function calling guide: https://developers.openai.com/api/docs/guides/function-calling
- Anthropic tool definition docs: https://platform.claude.com/docs/en/agents-and-tools/tool-use/define-tools
- Anthropic strict tool use: https://platform.claude.com/docs/en/agents-and-tools/tool-use/strict-tool-use
- Google Vertex AI function calling: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling

rxbynerd:
Completes issue #222 by adding the three reliability controls Waves 2–5 left unbuilt: **parallel-tool-call policy**, **input examples**, and **tool annotations**. (Tool-choice, strict mode, and structured results already shipped via #230/#228/#231 through other channels.)

## Design
- `types.ToolDefinition` gains a `Presentation *ToolPresentation` neighbour carried as `json:"-"`, so the nil zero value is byte-identical to the pre-#222 wire shape on every adapter — critical because the Anthropic adapter serialises `ToolDefinition` verbatim. `ToolPresentation` holds `InputExamples` and MCP-style `Annotations`.
- `StreamParams.ParallelToolCalls *bool` (nil = omit, mirroring `Temperature`).
- Two cross-provider quirks capabilities: `ParallelToolCallsCapability` and `ToolExamplesCapability`, gated per (provider, model) like the existing tool-choice/structured-result capabilities.

## Per-provider wire mapping
- **OpenAI Chat / Responses**: top-level `parallel_tool_calls`; examples folded into the JSON-Schema `examples` keyword **only for non-strict tools** (the structured-outputs subset rejects `examples`). Responses gains parallel + examples support (it previously had no tool_choice/parallel wiring).
- **Anthropic**: parallel-disable rides on a synthesised `tool_choice.disable_parallel_tool_use`; examples fold into `input_schema`.
- **Gemini / Bedrock**: deliberate no-op (no native control; Gemini's dialect rejects `examples`, so the #227 description text remains the carrier).
- **MCP**: server-declared annotations are now parsed and carried (advisory only — the conservative WorkspaceMutating/RequiresApproval gating still governs).

Built-in tool examples migrated from inline description text to structured `InputExamples`, pinned byte-for-byte to the descriptions by drift-guard tests. Contract tests cover every adapter; `provider-quirks.md` documents the per-provider matrix. `just test` + `just lint` green; the offline tool-use replay gate passes.

Closes #222.

### DIFF
--- a/harness/internal/provider/toolreliability_test.go
+++ b/harness/internal/provider/toolreliability_test.go
@@ -1,0 +1,283 @@
+package provider
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/rxbynerd/stirrup/harness/internal/provider/quirks"
+	"github.com/rxbynerd/stirrup/types"
+)
+
+// These tests pin the wire behaviour of the #222 reliability controls
+// (parallel-tool-call policy and input examples) per provider, gated on the
+// resolved quirks capability. They exercise the builders directly — no live
+// calls — mirroring the existing *_builder_test.go contract pattern.
+
+func toolWith222Example() types.ToolDefinition {
+	return types.ToolDefinition{
+		Name:        "demo",
+		Description: "Demo tool. Example: {\"x\": \"hi\"}",
+		InputSchema: json.RawMessage(`{"type":"object","properties":{"x":{"type":"string"}},"required":["x"]}`),
+		Presentation: &types.ToolPresentation{
+			InputExamples: []json.RawMessage{json.RawMessage(`{"x":"hi"}`)},
+		},
+	}
+}
+
+func userTurn() []types.Message {
+	return []types.Message{{Role: "user", Content: []types.ContentBlock{{Type: "text", Text: "go"}}}}
+}
+
+// decodeObject unmarshals raw JSON into a string-keyed map, failing the test
+// on error. Used to walk a wire body without committing to a typed shape.
+func decodeObject(t *testing.T, raw json.RawMessage) map[string]json.RawMessage {
+	t.Helper()
+	var m map[string]json.RawMessage
+	if err := json.Unmarshal(raw, &m); err != nil {
+		t.Fatalf("decode object: %v\nraw: %s", err, raw)
+	}
+	return m
+}
+
+func schemaHasExamples(t *testing.T, schema json.RawMessage) bool {
+	t.Helper()
+	_, ok := decodeObject(t, schema)["examples"]
+	return ok
+}
+
+func TestOpenAIChat_222_ParallelAndExamples_NonStrict(t *testing.T) {
+	disable := false
+	params := types.StreamParams{
+		Model:             "gpt-4o", // non-strict
+		Messages:          userTurn(),
+		Tools:             []types.ToolDefinition{toolWith222Example()},
+		MaxTokens:         100,
+		ParallelToolCalls: &disable,
+	}
+	q := quirks.DefaultRegistry().Resolve("openai-compatible", params.Model)
+	req, err := buildOpenAIRequest(params, true, q, nil)
+	if err != nil {
+		t.Fatalf("buildOpenAIRequest: %v", err)
+	}
+	body, err := json.Marshal(req)
+	if err != nil {
+		t.Fatalf("marshal: %v", err)
+	}
+	top := decodeObject(t, body)
+	if string(top["parallel_tool_calls"]) != "false" {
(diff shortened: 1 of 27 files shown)


harness/internal/provider/toolreliability_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"github.com/rxbynerd/stirrup/harness/internal/provider/quirks"
+	"github.com/rxbynerd/stirrup/types"
+)
+
+// These tests pin the wire behaviour of the #222 reliability controls
+// (parallel-tool-call policy and input examples) per provider, gated on the
+// resolved quirks capability. They exercise the builders directly — no live
+// calls — mirroring the existing *_builder_test.go contract pattern.
+
+func toolWith222Example() types.ToolDefinition {
```
rxbynerd, line 16: not a fan of including issue numbers inside the codebase, let alone in function names.

PR #363, opened 2026-05-28
types: add JSON marshaling and validation to ToolChoiceMode

### LINKED ISSUE #344: types: add IsValid, UnmarshalJSON, and MarshalJSON to ToolChoiceMode
Deferred from #322 (Wave 4 A1, #230 tool-choice plumbing).

## Item
`ToolChoiceMode` is an `iota` enum that currently lacks the safety/ergonomic methods its sibling enums (`OpenAITokenField`, `GeminiStreamArgsShape`) already have:

1. **`MarshalJSON`** — the integer form is opaque in trace/log JSON; a string form is self-documenting.
2. **`UnmarshalJSON` + `IsValid()`** — out-of-range integers are accepted silently. No external deserialization path exists today, so this is currently unexploitable.

## Files
- `types/events.go` (`ToolChoiceMode` const block)

## Rationale / when it becomes load-bearing
security-review LOW (A1), api-review §1. Consolidated because both touch the same enum and the same file. Becomes load-bearing when `StreamParams` (or any struct carrying `ToolChoiceMode`) gains an external deserialization path — e.g. a batch-replay or stored-RunConfig round-trip — at which point silent acceptance of an out-of-range value and opaque integer traces both become real defects.

rxbynerd:
`ToolChoiceMode` previously had no validation or custom JSON handling, so invalid values could round-trip silently through `RunConfig` and provider wiring. This adds `IsValid`, `String`, `UnmarshalJSON`, and `MarshalJSON` so the type rejects unknown values at the (de)serialization boundary and presents a consistent wire form.

Closes #344

### DIFF
--- a/types/events.go
+++ b/types/events.go
@@ -105,6 +105,77 @@ const (
 // ToolChoiceAuto fails to compile the moment ToolChoiceAuto is non-zero.
 var _ = [1]struct{}{}[ToolChoiceAuto]
 
+// String renders ToolChoiceMode as its stable lowercase wire form. These
+// are the same tokens MarshalJSON/UnmarshalJSON round-trip through, so a
+// trace or log line names the mode rather than an opaque integer. An
+// out-of-range value renders as "unknown(N)" — String never returns an
+// error, so it stays usable in fmt verbs; IsValid is the predicate that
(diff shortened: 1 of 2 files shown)


rxbynerd:
Duplicated work

Issue #343, opened 2026-05-25
provider: warn when tool-choice is silently dropped on unsupported adapters


rxbynerd:
Deferred from #322 (Wave 4 A1, #230 tool-choice plumbing).

## Item
Two provider adapters silently no-op a requested tool-choice mode instead of emitting an observable warning:

1. **`openai-responses`** — emit `slog.Warn` in `buildResponsesRequest` when `params.ToolChoice != ToolChoiceAuto && !q.ToolChoice.Supported`.
2. **`bedrock`** — emit `slog.Warn` in `BedrockAdapter.Stream` when `params.ToolChoice != ToolChoiceAuto`.

(Distinct from #332, which moves Responses wire divergences into quirks rules; this is purely an observability gap at the tool-choice drop site.)

## Files
- `harness/internal/provider/openai_responses.go` (`buildResponsesRequest`)
- `harness/internal/provider/bedrock.go` (`BedrockAdapter.Stream`)

## Rationale / when it becomes load-bearing
security-review LOW (A1) + api-review §4. An operator requesting `ToolChoiceRequired`/`ToolChoiceNone` against one of these adapters gets no signal that the escalation/required-tool guarantee is silently downgraded to auto. Becomes load-bearing once tool-choice escalation (#230) is enabled against a Responses or Bedrock run, where a dropped required-tool request looks like a model failure rather than an unsupported-capability degradation.

PR #379, opened 2026-05-28
provider: TODO the untested tool-choice suppression path

rxbynerd:
The #343 downgrade warn for openai-responses tool-choice is gated on
`!q.ToolChoice.Supported`, but no built-in quirk rule sets `Supported`
for openai-responses today, so the suppressed branch is unreachable and
untested — a reviewer flagged the gap (R3, low). This adds a `TODO(#343)`
in both the `Stream` guard and the warn-test cluster so the missing
suppression test gets wired in when the first openai-responses native
tool-choice quirk rule lands, rather than being forgotten.

Closes #0


rxbynerd:
Duplicate work

PR #418, opened 2026-06-07
permission, core: enforce the rule-of-two latch by revoking egress

rxbynerd:
## Summary

Wave 4 of 5 making the Rule of Two's sensitive-data leg dynamic (Ring 4). **Stacked on #417.** This is the behaviour-changing wave — the latch that #417 ships dark now *acts*.

- `harness/internal/permission/ruleoftwogate.go`: a latch-aware `PermissionPolicy` wrapper (with `Unwrap()`). While the run's monitor is untripped, or not enforcing, or the tool isn't egress-capable, it delegates to the inner policy unchanged. Once the latch trips under `block-external`, it denies every external-comm tool for the rest of the run.
- External-comm set: `run_command`, `web_fetch`, and any `mcp_`-prefixed tool — keyed on internal dispatch identity, so profile aliases can't dodge it.
- Actions (`ruleOfTwo.runtime.onDetect`): **`block-external`** (default — revoke egress, local work finishes), `ask-upstream` (route external calls through human approval; grpc-pinned by PR1's validator; inner policy consulted first so a Cedar `forbid` still denies), `redact` (rewrite matched latch-tier spans in tool-result `Content` **and** `Structured`, keep scanning post-latch), `abort` (terminal outcome `rule_of_two_violation`), `warn` (observe-only).
- **Default flip**: a run holding untrusted input + external comms with no declared sensitivity now auto-arms *enforcing*. The Wave-3 dark-ship pin test is inverted — the identical scenario now denies egress mid-run.
- Sub-agent latch sharing hardened: `permission.Unwrap` + `RewrapChain` rebuild the wrapper chain around the `ForChildRun` Cedar clone so a spawned child keeps both its `parentRunId` policy clone and the shared gate — `spawn_agent` is not a latch escape hatch.
- Metric `stirrup.ruleoftwo.actions{action}`; gate denials also ride the existing `permission_denied` event and `stirrup.permission.decisions` (no double-emit).

## Operator-visible behaviour change (read before merge)

Runs that previously held two-of-three freely and *also* ingested sensitive content at runtime will now **lose egress on detection** under the default posture. Escape hatches, both reviewable in the RunConfig:
- `ruleOfTwo.runtime.classifier: "none"` — disarm detection entirely.
- `ruleOfTwo.enforce: false` — disarm enforcement, keep auditable detection events.

Known false-positive footgun: Luhn-valid test PANs in fixtures (e.g. `4111111111111111`) trip the latch. FP cost here is availability (the run finishes locally without egress), not a safety failure.

## One-way decisions

- **Denial reason string** `rule_of_two: sensitive data observed in conversation; external communication revoked for this run` is a stable grep/audit key (`permission.RuleOfTwoDeniedReason`).
- **Abort outcome** `rule_of_two_violation` joins the terminal-outcome vocabulary.
- **Layering**: the gate consumes a local `permission.RuleOfTwoState` 3-method interface (`Tripped`/`Enforcing`/`Action`), not an import of `harness/internal/ruleoftwo` — importing the monitor into `permission` would invert leaf-package layering. Mirrors how `askupstream.go` declares its own `Transport` interface.
- **`onDetect: "abort"` + statically-declared sensitivity is rejected at validation** — a pre-tripped latch can never produce the runtime `Transition` that abort fires on, so the combination is dead config; the validator points the operator at `block-external` instead.

## Assumptions

- The gate is reached because every external tool is `WorkspaceMutating` or `RequiresApproval`, so it already routes through `Permissions.Check`. A future external tool that is neither would need adding to the set.
- `redact` operates on `DetectSensitive` span offsets over both the `Content` string and the separate `Structured` JSON string of each tool-result block; the operator `prompt` latches but is never rewritten (rewriting the task statement changes run semantics).

## Honest gap (unchanged, documented in Wave 5)

A single `run_command` that reads and exfiltrates in one shell pipeline — where the data never enters the conversation context — is not caught by this ring (it never reaches a scan boundary). This is Ring 2/3 territory and is the only same-turn exfil path; it predates this work and is unchanged.

## Review

One wave of four parallel reviews (code / security / spec-compliance / consistency → synthesized) + remediation before opening. The security review's lead conclusion: `block-external` is structurally sound — every built-in external tool and every statically-registered MCP tool is denied once the latch trips, and the sub-agent escape hatch is closed by `RewrapChain`. Fixed before opening:

- **Unscrubbed `permission_request.Input`** (security, newly load-bearing): the ask-upstream path forwarded the tool call carrying the latch-tripping secret to the control plane with `Input` unscrubbed. Transport `Emit` now scrubs `HarnessEvent.Input` additively (json-validity fallback), in both stdio and grpc, with existing transport tests confirmed unchanged.
- **`mcp_` prefix gap** in the gate's ask-upstream policy (a latent auto-allow the block-external path didn't have).
- **abort dead-code config** rejected at validation (above).
- Plus: redact JSON fallback via `json.Marshal`, deduplicated tool-set construction, and the now-false Wave-3 "observe-only" comments removed.

## Testing

`just` + `just lint` green; `-race` green on transport / permission / core / ruleoftwo / security. Covers: gate delegate-until-tripped, external-set + `mcp_` prefix, ask-upstream inner-first + fake-transport approval, `Unwrap`/`RewrapChain` chain integrity, redact span-preservation across `Content`+`Structured`, abort outcome incl. turn-0, the inverted default-flip pin, and an end-to-end stdio run where a seeded fake key trips the latch and the subsequent `run_command` is denied while the run completes cleanly.

### DIFF
--- a/harness/internal/core/factory.go
+++ b/harness/internal/core/factory.go
@@ -1102,9 +1102,9 @@ func editStrategyTool(es edit.EditStrategy, exec executor.Executor) *tool.Tool {
 // an LLM guard is expected to name when it spots sensitive content.
 var defaultRuleOfTwoGuardCriteria = []string{"sensitive_data", "pii"}
 
-// defaultRuleOfTwoAction is the documented onDetect default. Inert this
-// wave (no enforcement consumer); it flows into events so the soak
-// shows the action wave 4 will start applying.
+// defaultRuleOfTwoAction is the documented onDetect default, enforced
+// via the permission gate (block-external) when the arming matrix arms
+// the run.
 const defaultRuleOfTwoAction = "block-external"
 
 // ruleOfTwoArming is the factory's arming decision for the Rule-of-Two
@@ -1649,9 +1649,13 @@ func wrapRuleOfTwoGate(pp permission.PermissionPolicy, monitor ruleoftwo.Monitor
 	case "block-external":
 		return permission.NewRuleOfTwoGate(pp, monitor, externalCommToolSet(registry), nil, metrics)
 	case "ask-upstream":
+		// One external-comm set shared by the ask policy and the gate:
+		// both must cover the same tools, and a single allocation keeps
+		// them from silently diverging.
+		extSet := externalCommToolSet(registry)
 		timeout := time.Duration(cfg.PermissionPolicy.Timeout) * time.Second
-		ask := permission.NewAskUpstreamPolicy(tp, externalCommToolSet(registry), timeout)
-		return permission.NewRuleOfTwoGate(pp, monitor, externalCommToolSet(registry), ask, metrics)
+		ask := permission.NewAskUpstreamPolicy(tp, extSet, timeout)
+		return permission.NewRuleOfTwoGate(pp, monitor, extSet, ask, metrics)
 	default:
 		return pp
 	}
--- a/harness/internal/core/factory.go
+++ b/harness/internal/core/factory.go
@@ -1172,13 +1172,22 @@ func resolveRuleOfTwoArming(config *types.RunConfig) ruleOfTwoArming {
 			staticSensitive: s,
 		}
 	case classifier == "patterns":
+		// Explicit classifier:"patterns" with !u||!e is observe-only by
+		// construction (no enforcement consumer reads the latch here),
+		// so the monitor must NOT start pre-tripped even when the
+		// operator also declared sensitiveData: pre-tripping would skip
+		// every scan and yield zero detection telemetry — the opposite
+		// of what an operator who explicitly asked for pattern scanning
+		// wants. staticSensitive stays meaningful only on the enforcing
+		// u&&e path above, where a pre-tripped latch is the audited
+		// posture.
 		return ruleOfTwoArming{
 			armed:           true,
 			enforcing:       false,
 			action:          action,
 			criteria:        criteria,
 			classifier:      "patterns",
-			staticSensitive: s,
+			staticSensitive: false,
 		}
 	default:
 		return ruleOfTwoArming{}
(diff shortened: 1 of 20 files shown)


rxbynerd:
Rebased onto current `main` (`d34eb84`, ~125 commits) ahead of the v0.1 push. Same 4-commit stack, same 20-file scope.

**Conflicts and resolutions**

- `harness/internal/core/loop.go` (in `core, ruleoftwo, security: add redact and abort loop actions…`): collided with PR #462's `finalAssistantText` accumulator in `Run()`. Kept both — main's accumulator declaration and this branch's `turnZeroAbort → rule_of_two_violation` outcome flip. The mid-loop abort return also needed a third return value to match #462's new `runInnerLoop` signature (`return messages, "rule_of_two_violation", finalAssistantText`); folded into the same commit so every commit in the stack compiles.
- `harness/internal/core/types.go` (in `core: tidy rule-of-two gate construction…`): the `RuleOfTwo` doc-comment rewrite collided with the `Hooks`/`Shutdown` fields from the lifecycle-hooks PR (#467). Kept main's fields and comments, took this branch's updated enforcement-era comment.

**Preservation checks**

- `collectUntrustedChunks` retains main's `b.Structured` inclusion verbatim (the structured-content latch hole fix) — verified post-rebase.
- No new externally-communicating tools landed on main since the branch base (`externalCommToolSet` = `run_command`, `web_fetch`, `mcp_*` is still complete; the git builtins are read-only local).
- Turn-0 abort ordering vs lifecycle hooks: latch scan still precedes `Prompt.Build`; on abort the provider loop is skipped but `preRun`/`postRun` stay paired, with `postRun` seeing the `rule_of_two_violation` outcome.
- Sub-agent spawn still shares the parent monitor and rebuilds the gate wrapper chain around the per-child Cedar clone (composes with #467's `childConfig.Hooks = nil`).

**Local status**: `just build`, `just test`, `just lint` (0 issues) all pass; every commit in the stack compiles individually (`git rebase --exec go build`).

PR #419, opened 2026-06-07
docs, eval: document the rule-of-two runtime classifier and add deterministic suites

rxbynerd:
## Summary

Wave 5 of 5 (final) making the Rule of Two's sensitive-data leg dynamic (Ring 4). **Stacked on #418.** Documents what Waves 1–4 shipped and adds a deterministic eval suite. No production code changes.

- **`docs/safety-rings.md` Ring 4 rewrite**: the "What about the runtime path?" section, formerly an aspiration ("a future iteration may have GuardRail lift holdsSensitiveData"), is now the spec of the shipped runtime classifier — the deterministic detector, the one-way latch, the arming matrix, the `onDetect` action table, and the guard-criterion ratchet as one-way defence-in-depth. The "How they compose" scenario is rewritten so Ring 4 now trips `block-external` on runtime detection *without* a prior operator declaration. The summary table row and "what 'sensitive data' means" section updated to distinguish config-referenced secrets (not sensitive) from observed-content key sightings (now detected).
- **`docs/configuration.md`**: `ruleOfTwo.runtime` schema reference (each field, closed value set, default, the ask-upstream⇒grpc and abort+staticSensitive constraints).
- **`docs/guardrails.md`**: cross-reference to the criterion ratchet.
- **`eval/suites/ruleoftwo.hcl` + `ruleoftwo-observe.hcl`**: a fully deterministic suite (no vLLM/guard dependency) — secret-in-tool-result blocks egress, Luhn PAN in prompt blocks egress, AWS-docs EXAMPLE keys do not over-block, `enforce:false` keeps egress.

## Corrections made during implementation (worth noting)

- **The plan's own 50 KB-gap hypothesis was wrong, and is corrected in the docs.** The plan speculated about a "model-visible-but-unscanned" window past the dynamic-context cap. Verified against `loop.go` `Run()` and `dynamiccontext.go`: `SanitizeDynamicContext` truncates to 50 KB *before both* the turn-0 classifier scan and `Prompt.Build`, so the model and the scanner see the same truncated copy — no such window exists. The docs document the narrow real residual (the cap, not the classifier, bounds a single oversized entry; `read_file` content is uncapped) honestly.
- **Stale `enforce:false` rationale removed.** The canonical Dev/Defaults/Hardened configs carried `ruleOfTwo.enforce: false` justified by the `secret://` provider ref — false since #73 (`ruleOfTwoSensitiveData` never reads `apiKeyRef`). Removed from all three. Dev/Defaults have no untrusted-input leg, so all-three never held. **Hardened is now the runtime-classifier showcase**: it has `web_fetch` (untrusted + external both hold), so dropping the override lets the factory auto-arm enforcing `block-external` — a fetched page delivering secret/PII content revokes egress mid-run.

## One-way decisions

- **Hardened canonical config now demonstrates enforcing `block-external`** rather than carrying a (previously dead) override — this is the reference posture operators will copy.
- The eval suite is split into two files (`ruleoftwo.hcl` enforcing + `ruleoftwo-observe.hcl` observe) because `LoadSuiteHCL` is one-suite-per-file and `rule_of_two` is not a per-task override.

## Assumptions / documented limitations

- **Eval determinism**: egress-blocked tasks assert the exfil sentinel is *absent* (LLM variance fails safe — fully deterministic outcome). Over-block and enforce:false tasks assert sentinel *present* (best-effort positives, depend on the agent running a benign command). "Detection events fire under enforce:false" is not expressible in the eval judge and is documented in-comment, covered instead by the Wave-3/4 unit + integration tests.
- The suites carry no `../baselines/` entry (like `tooluse.hcl`), so CI's compare step skips them; they need `ANTHROPIC_API_KEY` to run and are skipped otherwise.
- **GuardToolCall fires before the gate**: the suites deliberately use non-credential filenames (`deploy-notes.txt`, not `.env`) and benign commands (`ls`, not `curl`) so they exercise the rule-of-two gate specifically rather than the pre-existing tripwire that would mask it.

## Honest gaps documented (in "What these don't catch")

- A single `run_command` that reads and exfiltrates in one shell pipeline, where data never enters the conversation context — Ring 2/3 territory, unchanged by this work, the only same-turn exfil path.
- The 50 KB dynamic-context cap residual described above.

## Review

One doc-focused review wave before opening: `doc-tone-reviewer` (impersonal-voice convention — 4 fixes applied: eliminated "you can hold", two "Use this when", a colloquial "must live in") and `technical-doc-reviewer` (accuracy against the shipped code). The technical review confirmed no doc claim is wrong — schema, arming matrix, action behaviours, denial string, and all three metric names are exact matches — and independently re-verified the 50 KB-truncation correction and the stale-rationale fix against source.

## Testing

`go test ./eval/...` green; `stirrup-eval run --suite … --dry-run` loads both suites and every merged config passes `ValidateRunConfig` (3/3 and 1/1 tasks); new `eval/spec` parse test pins the `rule_of_two { runtime { ... } }` grammar; all safety-rings.md JSON blocks parse; internal doc anchors resolve.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -139,6 +139,7 @@ Quick map for "I need to change X" lookups:
 | Lifecycle hooks (pre/post-run exec, #461) | `harness/internal/hook/` (Runner/Noop/ExecRunner), loop wiring in `harness/internal/core/loop.go` (operator doc: `docs/configuration.md#lifecycle-hooks`) |
 | Permission gating logic | `harness/internal/permission/<type>.go` |
 | Cedar policy semantics | `harness/internal/permission/policyengine.go` |
+| Rule-of-Two runtime classifier (detector / monitor+latch / gate) | `harness/internal/security/sensitivepatterns.go`, `harness/internal/ruleoftwo/`, `harness/internal/permission/ruleoftwogate.go` (arming in `core/factory.go`; operator doc: `docs/safety-rings.md`) |
 | Container runtime / network mode wiring | `harness/internal/executor/container*.go` |
 | K8s executor (Pod-per-run) + egress NetworkPolicy | `harness/internal/executor/k8s.go`, `k8s_netpol.go` (operator doc: `docs/executors/k8s.md`) |
 | `k8s-sandbox` executor (Agent Sandbox CRD: `agents.x-k8s.io/v1alpha1` Sandbox → controller-created Pod) | `harness/internal/executor/agentsandbox.go`, `agentsandbox_api.go` (operator doc: `docs/executors/k8s-agent-sandbox.md`) |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -792,10 +792,10 @@ versions.
 ## Rule-of-Two configuration
 
 The `ruleOfTwo` block tunes [Ring 4](safety-rings.md#ring-4--rule-of-two-pre-flight-invariant--runtime-classifier).
-It has no CLI flag — the override must live in the RunConfig so it is
+It has no CLI flag — the override must be stored in the RunConfig so it is
 reviewable in pull requests. The behavioural semantics (what trips the
 latch, the arming matrix, the on-detect actions) are documented in
-[`safety-rings.md`](safety-rings.md#the-runtime-classifier); this is the
+[`safety-rings.md`](safety-rings.md#the-runtime-classifier); this table is the
 schema reference.
 
 | Field | Type | Default | Notes |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -789,6 +789,39 @@ Use `prompt.model` / `prompt.tier` (and `gen_ai.system_instructions`
 under `--otel-capture-content`) to correlate runs with prompt
 versions.
 
+## Rule-of-Two configuration
+
+The `ruleOfTwo` block tunes [Ring 4](safety-rings.md#ring-4--rule-of-two-pre-flight-invariant--runtime-classifier).
+It has no CLI flag — the override must live in the RunConfig so it is
+reviewable in pull requests. The behavioural semantics (what trips the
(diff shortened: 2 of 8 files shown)


rxbynerd:
Rebased onto the freshly rebased `feat/rule-of-two-enforcement` tip (`f2f9aa1`, itself now on `main`@`d34eb84`). No conflicts — all 5 commits replayed cleanly.

**Docs cross-checked against post-rebase main**

- Main's lifecycle-hooks additions to `safety-rings.md` (the "Lifecycle hooks are outside Rings 3, 4, and 5" scope bullet from #467's doc pass) survived the Ring 4 rewrite and remain consistent with the classifier framing (hooks add no agent-reachable capability; the gate keys on tool IDs).
- Anchor consistency with the open #482: this branch keeps the `## Ring 3 — Cedar policy engine (per-call authorization)` heading intact, so #482's two `guardrails.md` anchor fixes still resolve after either merges. All anchors this branch adds (`#the-runtime-classifier`, `#ring-4--rule-of-two-pre-flight-invariant--runtime-classifier`, `#the-guard-criterion-ratchet`) resolve against the rewritten headings; no other doc links to the removed `#what-about-the-runtime-path` / old Ring 4 anchor.
- `configuration.md` schema claims re-verified against code: the three `validateRuleOfTwoRuntime` cross-field rules, the `["sensitive_data", "pii"]` factory default, the `rule_of_two_disabled` event, and the never-inject-runtime-block behaviour all match.

**Eval suites**: both new suites validate end-to-end — `stirrup-eval run -dry-run` passes for `eval/suites/ruleoftwo.hcl` (3 tasks) and `eval/suites/ruleoftwo-observe.hcl` (1 task); `eval/spec` parser tests pass.

**Local status**: `just build`, `just test`, `just lint` (0 issues) all pass; every commit compiles individually.

Issue #310, opened 2026-05-24
observability: bound provider.model metric label across all provider types


rxbynerd:
Cardinality follow-up from #304's `stirrup.harness.tool_failures` counter (and the four sibling tool metrics that share the same label set).

## Issue

`ValidateRunConfig` enforces `^[A-Za-z0-9._-]{1,64}$` on `ProviderConfig.Model` only for the Gemini provider. For `anthropic`, `openai-compatible`, `openai-responses`, and `bedrock`, the model string has no character-class or length bound. The `provider.model` label flows into:

- `harness/internal/core/dispatch.go:291`
- `harness/internal/core/loop.go:667,727`

…all of which feed OTel metrics. An operator with `RunConfig` authoring power (Cloud Run job spec, gRPC submitter) can set an unreasonably long or non-printable model string and inflate TSDB cardinality / break dashboard label rendering.

## Threat model

Exploitation requires attacker control of `RunConfig`, which is operator-level. Blast radius is limited: this is configuration self-harm, not a model-controlled vector like the `tool.name` issue that #304 hardened.

## Fix options

1. Extend `observabilityLabelPattern` validation to `Model` for all provider types in `validateProviderConfigs` (preferred — fail-closed at config time, matches the Gemini precedent).
2. Sanitise via `sanitiseLabel` at the metric emission boundary as defence-in-depth (does not protect dashboards configured from raw RunConfig).

Option 1 is the load-bearing fix; option 2 is optional belt-and-braces.

## Why deferred

Security-reviewer rated this LOW; not a blocker for #304. Per the wave-1-issue-229 synthesis.

Followed up from #304.

PR #420, opened 2026-06-07
types: allow forward slashes in model names for OpenRouter support

rxbynerd:
Create a separate modelLabelPattern regex that allows forward slashes in model names, enabling valid OpenRouter-style names like "deepseek/deepseek-v4-flash" and "anthropic/claude-sonnet-4.6" to pass validation. The original observabilityLabelPattern remains restrictive for actual OTel resource labels (environment, service.namespace) where slashes are not appropriate.

This change keeps the cardinality guard from #310 while supporting provider namespacing in model identifiers. Bedrock model ARNs with slashes continue to work via their existing exemption.


rxbynerd:
Haiku PR

PR #443, opened 2026-06-30
Support Qwen 3.6 via LM Studio (local OpenAI-compatible)

rxbynerd:
## Summary

Adds support for running a locally-hosted **Qwen 3.6** model through **LM Studio**'s OpenAI-compatible endpoint. The work was driven by probing a live LM Studio server and running a real agentic loop end-to-end, cross-checked against a cited research packet.

**Headline finding: Qwen 3.6 needs no wire-shape quirks — it is the simplest of the reasoning families.** Unlike DeepSeek v4 and Z.ai GLM, it does **not** return HTTP 400 when prior-turn `reasoning_content` is omitted (verified 200), accepts the modern `max_completion_tokens` key, and welcomes sampling parameters. So there is **no `ReplayFields` rule, no `TokenField` override, no `OmitSamplingParams`, and no compat profile**. LM Studio extracts the chain-of-thought into a `reasoning_content` field that the harness correctly drops between turns — matching Qwen's own chat template.

What actually blocked "full potential" were two **general `openai-compatible` correctness bugs** that Qwen happens to expose, plus operator guidance. The three commits are independent and split cleanly.

### 1. `provider: account streaming output tokens via stream_options.include_usage`
The adapter never sent `stream_options` and read output usage only from a block on the `finish_reason` chunk. OpenAI itself and every spec-compliant compatible server (LM Studio, vLLM, DeepSeek) emit **no usage in a stream** unless `stream_options.include_usage` is set, and deliver it in a **trailing chunk whose `choices` array is empty** — arriving *after* the `finish_reason` chunk that already produced `message_complete`. Result: output tokens resolved to **0** for every streaming `openai-compatible` turn (input is unaffected because it is locally estimated, not read from the stream).

Fix: request `stream_options.include_usage` on streaming turns and capture the usage block wherever it lands, surfacing a late count as a usage-only `message_complete`. Empty stop reason / replay fields on that event keep `streamEventsToResult`'s last-write merge from clobbering the real values. The six streaming request goldens gain the new field; a Qwen 3.6 SSE capture pins the trailing-usage path and the deliberate absence of a `reasoning_content` replay rule.

**Verified live:** a run that previously reported `0 out` now reports the true count (`272 out`).

### 2. `quirks: advertise openai-compatible capabilities for vendor-prefixed ids`
`path.Match`'s `*` does not cross `/`, so the universal `openai-compatible` `tool_choice` and `parallel_tool_calls`/examples base rules (`ModelMatch: "*"`) never matched the `vendor/model` ids that LM Studio and OpenRouter serve (`qwen/qwen3.6-27b`, `deepseek/deepseek-v4-flash`, `mlx-community/...`). A bare id resolved the full tool surface; the same model under a one-level prefix silently resolved to the zero (unsupported) value. The DeepSeek gateway rule documented this and explicitly deferred it as "a separate decision" — this makes that decision.

Adds `*/*` sibling rules beside each universal capability rule (the two globs are disjoint, so there is no ordering interaction; deeper nesting is intentionally unmatched), and updates the DeepSeek Rule D comment + its resolution test to reflect that gateway ids now advertise the same tool surface as bare ids, with the DeepSeek-specific wire behaviour still coming solely from its own rule.

**Verified:** `stirrup providers quirks --provider openai-compatible --model qwen/qwen3.6-27b` now resolves `tool_choice`, `parallel_tool_calls`, and `examples` as supported, while `replayFields` stays empty.

### 3. `docs: document Qwen 3.6 via LM Studio and add an example runconfig`
Adds an operator-facing guide (`docs/providers.md` → "Local models via LM Studio (Qwen 3.6)") and a ready-to-edit `examples/runconfig/qwen3.6-lmstudio.json`. The guide records the load-bearing operational details: thinking is always on (neither `/no_think` nor `enable_thinking` is reachable via the plain Chat Completions surface), `temperature` should be raised to Qwen's recommended **0.6** (the harness default 0.1 is below spec and greedy decoding is discouraged), and vendor-prefixed ids now resolve the full tool surface.

## ⚠️ Local-model footgun found and documented (no code change here)

While driving it live I root-caused a confusing `empty stop reason` failure: `contextStrategy.maxTokens` must stay **above the harness's hardcoded 64k response reserve** (`defaultReserveForResponse`, `core/loop.go`). Below it, `available = maxTokens - 64000 <= 0`, the sliding window truncates to the **last two messages every turn** (dropping the user message), and the model returns a malformed turn. My example first used `32768` and failed every run; it now uses `131072` and is verified end-to-end. The docs call this out.

The deeper limitation — small-context LM Studio deployments (≤64k) get no usable prompt budget because the 64k reserve is fixed — is **deliberately left out of this PR**. A hard validation would break existing `harness_test.go` cases that use `MaxTokens: 1000` to exercise truncation. **Suggested follow-up:** make the response reserve / output budget configurable (or adaptive to `contextStrategy.maxTokens`), and warn at loop construction when `maxTokens <= reserve`.

## Test plan
- `just test` and `golangci-lint run ./...` green.
- New tests: trailing-usage SSE parse, Qwen "no wire quirks" request shape, and the `*/*` capability siblings (incl. the one-prefix-level boundary).
- **End-to-end against a live LM Studio (Qwen 3.6 27B):** agentic edit task completes (`success`), tool calling works through the loop, and output-token accounting is correct (`0 → 272`).

## Review notes
The two general fixes (commits 1 and 2) are not Qwen-specific and could land independently of the docs commit if you'd prefer to split this into separate PRs.

https://claude.ai/code/session_017ns5rVNZY5rg21ZRTs12b1


rxbynerd:
Filed the context-budget footgun called out in this PR's description as #444 (decoupling the provider output budget from the 64k context reserve so small-context local models work). The example config + docs here keep `contextStrategy.maxTokens` above 64k as the interim workaround.

Issue #164, opened 2026-05-10
packaging/deployment: Google Cloud Run job


rxbynerd:
## Background

Stirrup is primarily a Kubernetes-native job (`stirrup job` + control-plane gRPC), but the binary is already shaped as a finite, exit-on-completion process — exactly what Cloud Run **jobs** were designed to host. With #119 (Vertex AI Gemini adapter) merged and #162 (Vertex Gemini WIF smoke test) and #163 (dual-publish container to GHCR + Google Artifact Registry via Direct WIF) in flight, the missing piece for a fully-GCP-native execution path is the deployment + invocation surface for Cloud Run jobs — and, critically, **a way to get the run's results back out**, since Cloud Run jobs (like every serverless target) have no native output channel.

The end-to-end target is: a GHA workflow that dispatches a Cloud Run job in `rubynerd-net`, the container pulled from `europe-west4-docker.pkg.dev/rubynerd-net/stirrup/stirrup`, the attached service account inheriting `roles/aiplatform.user` so the harness can talk to Vertex AI Gemini via the GCE metadata server with **no static credentials anywhere in scope**, and the run's results delivered back to the caller via two complementary mechanisms (one for direct consumption, one for artifact production).

### Title clarification: "function" vs "job" vs "service"

The issue title says *Cloud Run function* but the linked quickstart is to Cloud Run **jobs**. This matters because the three Cloud Run product surfaces have materially different container contracts:

| Surface | Container contract | Fit for stirrup |
|---|---|---|
| Cloud Run **service** | Must listen on `0.0.0.0:$PORT` for HTTP. Long-running, autoscaled. | **Bad fit.** Stirrup has no HTTP surface; wrapping it in one is busywork. |
| Cloud Run **functions** (2nd-gen Cloud Functions) | Runs on Cloud Run services under the hood — same HTTP requirement plus a Functions Framework wrapper. Event-driven. | **Bad fit.** Same HTTP problem with extra packaging overhead. |
| Cloud Run **jobs** | One-shot batch; container runs to completion, exit code is the result. No port. | **The right fit.** Matches stirrup's existing exit semantics. |

This issue therefore scopes to **Cloud Run jobs** (matching the linked quickstart) and treats the title wording as imprecise. *Cloud Run functions* would require a bespoke HTTP-front entrypoint and is an orthogonal future issue.

## Goal

[...]

Issue #461, opened 2026-07-02
control-plane: sandbox environment setup phase (repo clone + runtime provisioning before the agentic loop)


rxbynerd:
Surfaced while triaging the control-plane follow-ups ([`docs/CONTROL_PLANE.md`](docs/CONTROL_PLANE.md)). This is a gap §4.3 only half-addresses, and one that was not framed against the `local` executor before.

## Problem
Before the agentic loop can do useful work, the sandbox usually needs to be *set up*: the target repository cloned, and a runtime provisioned to the projects requirements (e.g. switch to a specific Ruby via `rbenv`/`asdf`, `bundle install`, run a project bootstrap script). Stirrup has no way for a control plane to instruct it to perform this setup — the harness assumes a ready workspace and starts turn 1 immediately.

This was invisible with the `local` executor on a developer machine (the human sets up their own checkout). It becomes load-bearing the moment stirrup runs:
- **`local` executor inside a Kubernetes Job** — the Pod *is* the sandbox, but nothing populates it. Who clones the repo and installs the Ruby version *inside that Pod* before stirrup starts?
- **`container` / `k8s` / `k8s-sandbox` (separate sandbox)** — tools run in a *different* container/Pod from the harness. An init container on the *harness* Pod can clone into a shared volume (filesystem), but it **cannot run runtime provisioning inside the tool sandbox** (version managers, `bundle install`, env). That environment is only reachable through `Executor.Exec` — which today only the agent loop drives.

## Current state (verified at `adac2b2`)
- No `setup`/`clone`/`bootstrap`/`prelude` field exists in `RunConfig` (`types/runconfig.go`).
- `GitStrategy.Setup` (`harness/internal/git/deterministic.go:25`) only creates a branch **inside an existing checkout** — it does not clone. §4.3 confirms "nothing in stirrup clones repositories."
- §4.3 pushes all materialisation to the control plane: init-container git clone, snapshot/tarball restore, PVC/volume mount, or image-baked sources.
- **But** the `Executor` interface (`harness/internal/executor/executor.go:29`) already provides `Exec`, `WriteFile`, and `Capabilities{CanExec, CanNetwork}` across every executor (`local`, `container`, `k8s`, `k8s-sandbox`) — a ready substrate for a setup phase that runs *where the tools run*.

## Options considered

**A. Control-plane-orchestrated only (status quo, §4.3).** Init container clones into a shared volume; a custom sandbox image bakes the runtime; PVC/snapshot restore.
[...]

PR #462, opened 2026-07-02
core, cmd: populate RunResult.finalAssistantText end-to-end

rxbynerd:
## Summary

Populates `RunResult.FinalAssistantText`, which was declared on the RunResult wire type but never filled in. Resolves CONTROL_PLANE.md §7 item 3. After every run the stdout-json result sink now carries the model's last non-empty assistant answer alongside the bookkeeping (turns/tokens/outcome), and the same value is persisted on the trace surfaces used by serverless/Cloud Run collection.

The loop already computed `lastAssistantText(sr.Blocks)` on the `end_turn` path but discarded it; this threads that value through to the trace and the RunResult.

## What changed

- `types/runtrace.go` — new `FinalAssistantText` field on `RunTrace` (omitempty).
- `harness/internal/core/loop.go` — captures the last non-empty assistant text across every turn (computed once, reused by the `end_turn` PhasePostTurn guard), and stamps it onto the finished trace at the single shared finalisation point. The value is:
  - **gated behind the post-turn guard**: a turn whose text the PhasePostTurn guard denies (`guardrail_blocked`) never appears in the field — the prior guard-approved value (or empty) is returned instead, so the new forwarding channel honours the guard's do-not-forward contract.
  - **scrubbed** via `security.Scrub` at a single site, consistent with every other externally-forwarded string in the loop.
- `harness/internal/trace/*` — new optional `FinalAssistantTextRecorder` capability interface (mirroring the existing `SystemInstructionsRecorder` pattern), implemented by all four emitters (jsonl, gcs, otel, nested_jsonl) so the persisted `run_finished` trace / GCS object carries the field, not just the in-memory RunResult. On the OTel emitter the value feeds only the returned in-process `RunTrace` aggregate (→ RunResult), not an exported span attribute, so it is intentionally independent of `captureContent` and does not affect OTel content export.
- `harness/cmd/stirrup/cmd/root.go` — `buildRunResult` maps the field onto `RunResult`; removed a stale stub comment.

Scope is contained to types + loop + trace + root + tests. No proto / gRPC `done` event changes; `buf generate` was not run.

## Tests

- `harness/internal/core/loop_test.go` — populated case (incl. JSONL-buffer re-parse proving the persisted trace carries it), last-non-empty-across-turns, not-clobbered-by-a-later-empty-turn, single-turn and two-turn guardrail-blocked no-leak, and a scrub assertion.
- `harness/cmd/stirrup/cmd/root_test.go`, `harness/internal/resultsink/resultsink_test.go` — mapping + sink round-trip, including omitempty (key absent when empty).

## Review

Implemented, then run through a reviewer wave (code, spec-compliance, test-coverage) and remediated before this PR: fixed a guardrail-denied-text leak through the new channel, closed a gap where the field never reached the persisted JSONL/GCS trace, added a secret-scrub pass matching the surrounding code, and added the regression tests above. `just build`, `just test`, and `just lint` all pass.

## Follow-up (not in this PR)

- No size bound / truncation on the field yet — a very long answer could truncate the `STIRRUP_RESULT` JSON line in Cloud Logging. Worth a documented byte cap in a follow-up.
- `docs/CONTROL_PLANE.md` §2.5 still says the field is "currently never populated — issue #164". That doc does not exist on `main` (it lives on the in-flight `control-plane` branch), so the row edit + stale-#164 removal must be applied there. `docs/configuration.md` / `README.md` have no such claim.

### DIFF
--- a/harness/cmd/stirrup/cmd/root.go
+++ b/harness/cmd/stirrup/cmd/root.go
@@ -136,12 +136,6 @@ func hookSummaryCounts(results []types.HookExecution) (ran, failed int) {
 // completed RunTrace. The mapping is straightforward; see
 // types/result.go for the schema's stability rules.
 //
-// FinalAssistantText is left empty for v1: the loop computes it (see
-// loop.go::lastAssistantText) but does not thread it onto RunTrace, and
-// the brief is explicit about keeping schema additions to Chunk A.
-// TODO(#164): expose the loop's last assistant text so the resultSink
-// can carry the run's *answer* as well as its bookkeeping.
-//
 // VerifierVerdict is populated from the most recent VerificationResult
 // when one exists. Treating an empty Feedback string as "no verifier
 // ran" would conflate "verifier passed silently" with "verifier was
@@ -160,12 +154,13 @@ func buildRunResult(rt *types.RunTrace) types.RunResult {
 		return types.RunResult{SchemaVersion: 1, Outcome: "internal-error"}
 	}
 	res := types.RunResult{
-		SchemaVersion: 1,
-		RunID:         rt.ID,
-		Outcome:       rt.Outcome,
-		Turns:         rt.Turns,
-		TokenUsage:    rt.TokenUsage,
-		DurationMs:    rt.CompletedAt.Sub(rt.StartedAt).Milliseconds(),
+		SchemaVersion:      1,
+		RunID:              rt.ID,
+		Outcome:            rt.Outcome,
+		Turns:              rt.Turns,
+		TokenUsage:         rt.TokenUsage,
+		DurationMs:         rt.CompletedAt.Sub(rt.StartedAt).Milliseconds(),
+		FinalAssistantText: rt.FinalAssistantText,
 	}
 	if n := len(rt.VerificationResults); n > 0 {
 		last := rt.VerificationResults[n-1]
--- a/harness/cmd/stirrup/cmd/root_test.go
+++ b/harness/cmd/stirrup/cmd/root_test.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"strings"
 	"sync"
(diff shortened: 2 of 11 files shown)


rxbynerd:
Rebased onto current `main` (9102b10, 32 commits ahead of the old base) with `--force-with-lease`. The PR's six commits are preserved 1:1; an interdiff of old-diff vs new-diff shows the only delta is a gofmt realignment of the `hookResults` field in `jsonl.go` (the new `finalAssistantText` field is longer).

## Conflicts and resolutions

All conflicts were both-additive collisions with the lifecycle-hooks feature (#467 / issue #461), which landed in the same files:

- **`types/runtrace.go`** — main rewrote the `Outcome` doc comment (adding `setup_failed` / `hook_failed`) and added `HookResults` + the `HookExecution` type; this PR adds `FinalAssistantText`. Kept main's comment block and hook types intact; inserted `FinalAssistantText` between `Outcome` and `HookResults`.
- **`harness/internal/trace/trace.go`** — both sides append an optional capability interface. Kept both: `HookRecorder` (main) and `FinalAssistantTextRecorder` (this PR).
- **`harness/internal/trace/jsonl.go`** — three hunks: the emitter struct (kept `hookResults` and `finalAssistantText`), the `Start()` reset block (reset both), and the `Finish()` `RunTrace` literal (carries both `FinalAssistantText` and `HookResults`). `RecordHookExecution` and `RecordFinalAssistantText` both retained.
- **`harness/cmd/stirrup/cmd/root_test.go`** — main added the hook-failure and shutdown-watchdog tests in the same region as `TestBuildRunResult_FinalAssistantText`. Kept all of main's tests; placed the FinalAssistantText test with the other `TestBuildRunResult_*` tests.

`loop.go`, `root.go`, `gcs.go`, `otel.go`, `nested_jsonl.go`, `loop_test.go`, and `resultsink_test.go` auto-merged cleanly. Verified the merged `Run()` ordering by hand: preRun hooks → inner loop accumulates the text → postRun hooks (may override outcome to `hook_failed`) → `stopHeartbeat` → scrub → `RecordFinalAssistantText` → `Trace.Finish`, so the recorder still fires before the emitters serialise the trace and hook outcome overrides are unaffected.

## Local verification

- `just build`, `just test`, `just lint` — all green (0 lint issues).
- Fake-key end-to-end run (`--output json`): `STIRRUP_RESULT` line emits and parses; the error path omits `finalAssistantText` via `omitempty` as designed (no assistant turn produced text).

No scope change: the length-cap follow-up remains #463.

PR #467, opened 2026-07-09
core, types, executor: lifecycle hooks — pre/post-run sandbox exec (#461)

### LINKED ISSUE #461: control-plane: sandbox environment setup phase (repo clone + runtime provisioning before the agentic loop)

rxbynerd:
Closes #461.

## What

`RunConfig` gains a `hooks` section: ordered, operator-authored shell commands executed in the sandbox through the run's `Executor` — `hooks.preRun` before turn 1 (repo clone, runtime provisioning, `bundle install`), `hooks.postRun` after the session (artifact submission, smoke tests). Deterministic setup/teardown stops consuming turns and tokens, and setup becomes expressible over the gRPC `task_assignment` path (proto field 32).

```json
{
  "hooks": {
    "preRun":  [{ "name": "clone", "command": "git clone https://… .", "timeoutSeconds": 300 }],
    "postRun": [{ "name": "smoke", "command": "bundle exec rake smoke", "runOn": "success" }]
  }
}
```

## Design decisions

- **Naming is per-run, not per-session** — `docs/sessions-spec-draft.md` reserves "session" for multi-run continuity.
- **Hook output is trace-only**: recorded as `hook_record` JSONL events and `RunTrace.HookResults` (scrubbed, 4 KB rune-safe tails), never injected into model context — no new prompt-injection surface.
- **Hooks share the run's network posture** (same egress allowlist / NetworkPolicy); no phase-specific egress.
- **Failure semantics**: fatal preRun failure → outcome `setup_failed`, session never starts; fatal postRun failure → `hook_failed`, only ever overriding `success` (a primary failure cause is never masked). Per-hook `continueOnError` and postRun `runOn: always|success|failure` for best-effort hooks.
- **postRun survives the run deadline but not process signals**: detached, budget-bounded context (≤1830 s) that still observes SIGTERM/SIGINT via an injected shutdown context plus a 5 s teardown watchdog — preserving the documented Cloud Run SIGTERM-grace contract and K8s sandbox teardown.
- **Read-only modes allow hooks** (consistent with test-runner verifier and deterministic git strategy precedent; documented in `docs/configuration.md` and pinned by test). **Rule-of-Two flags are untouched** by hooks (pinned by test).
- **Executor exec hard cap raised 5→30 min** for cold installs; the agent-reachable `run_command` path stays clamped at 300 s in the tool layer, pinned by a regression test.
- `secret://` is rejected (case-insensitively) inside hook commands; clone credentials arrive via control-plane runtime bindings per CONTROL_PLANE.md §2.7.

## Notable fixes shaken out along the way

- Fatal `Run()` errors carrying a trace (now including `setup_failed`) previously skipped `RunResult`/result-sink emission and the gRPC `done` event on both CLI entrypoints — fixed; `error` outcomes now also reach `done`.
- `LocalExecutor` exec left stdout pipes open to orphaned grandchildren after ctx cancellation, blocking `Exec` for the grandchild's lifetime — fixed via `cmd.WaitDelay` (SIGTERM response measured 27 s → 0.6 s).
- UTF-8 rune-boundary-safe truncation for persisted hook output tails.

## Verification

- Plan → implementation → 5-reviewer wave (code, security, spec-compliance, test-coverage, consistency) → synthesized remediation (14 findings fixed, 2 blocking) → fresh-context end-to-end verification of 8 runtime scenarios against the built binary, including live SIGTERM timing and `setup_failed` result emission.
- `just build`, `just test` (41 packages, `-race` clean on `harness/...`), `just lint` all green; new hook package at 94 % coverage.
- Docs: `configuration.md` (schema + worked example), `architecture.md`, `security.md`, `safety-rings.md` (hooks sit outside Rings 3/4/5), `cloud-run-jobs.md`, `executors/k8s.md` (`terminationGracePeriodSeconds`), `deployment.md`.

Follow-up issues for the six deferred review findings plus two verification observations are being filed and will be linked here.

https://claude.ai/code/session_01Gxw8GWTYdFraZWjmPk2SGt

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -33,6 +33,7 @@ stirrup/
       verifier/              # Verifier: none, test-runner, composite, llm-judge
       permission/            # PermissionPolicy: allow-all, deny-side-effects, ask-upstream, policy-engine (Cedar)
       git/                   # GitStrategy: none, deterministic
+      hook/                  # Lifecycle hooks: Runner, Noop, ExecRunner (pre/post-run exec, #461)
       transport/             # Transport: stdio, gRPC bidi streaming, null (sub-agents)
       guard/                 # GuardRail: none, granite-guardian, cloud-judge, composite, phase-gated
       trace/                 # TraceEmitter: JSONL, OpenTelemetry (OTLP/gRPC or OTLP/HTTP)
@@ -190,10 +191,14 @@ The `guard` package (`guard/`) provides an LLM-based safety classifier called at
 - `max_tokens: 64000`, `temperature: 0.1`
 - File size limit: 10MB (read/write)
 - Command output cap: 1MB
-- Command timeout: 30s default, 5min max
+- Command timeout: 30s default, 5min max (`run_command` tool clamp;
+  the shared `Executor.Exec` hard cap is 30min, raised for lifecycle
+  hooks — #461)
 - Follow-up grace cap: 3600s
 - Token budget cap: 50M
 - Cost budget cap: $100
+- Lifecycle hook timeout: 300s default, 1800s (30min) max per hook;
+  32 hooks max per phase
 
 ## Development
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -135,6 +135,7 @@ Quick map for "I need to change X" lookups:
 | Provider behaviour or wire format | `harness/internal/provider/<name>.go` |
 | Tool definition or schema | `harness/internal/tool/builtins/<name>.go` |
 | Edit fallback logic | `harness/internal/edit/multi.go` |
+| Lifecycle hooks (pre/post-run exec, #461) | `harness/internal/hook/` (Runner/Noop/ExecRunner), loop wiring in `harness/internal/core/loop.go` (operator doc: `docs/configuration.md#lifecycle-hooks`) |
 | Permission gating logic | `harness/internal/permission/<type>.go` |
 | Cedar policy semantics | `harness/internal/permission/policyengine.go` |
 | Container runtime / network mode wiring | `harness/internal/executor/container*.go` |
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -127,6 +127,24 @@ the `Verifier` validates output. The `Transport` carries events to
 and from the control plane; the `TraceEmitter` records spans and
 metrics throughout.
 
+Lifecycle hooks (issue #461) run around this loop, not inside it. A
+configured `hook.Runner` (`hook.Noop` when the run has none) executes
+`preRun` hooks after `PromptBuilder` builds the system prompt and
+before `GitStrategy.Setup` — the deterministic git strategy assumes an
+existing checkout, so a clone hook has to create it first — under the
+run's own wall-clock context. `postRun` hooks execute after outcome
+classification and `GitStrategy.Finalise`, before the run reports
+`done`, on a context detached from the run's own cancellation and
+deadline (`context.WithoutCancel`) and bounded by the sum of the
+configured `postRun` timeouts plus a 30 s margin, so an artifact
+upload can still finish after the run's wall-clock timeout has
+expired. Hook output never enters the model's context; it is
+trace-only, recorded via the optional `trace.HookRecorder` capability
+the same way `trace.SystemInstructionsRecorder` is. Sub-agent loops
+always get `hook.Noop` — hooks are parent-run-only. Full schema,
+defaults, and failure-semantics reference: [`configuration.md`
+"Lifecycle hooks"](configuration.md#lifecycle-hooks).
+
 ## Modes
 
 Five run modes ship as partial `RunConfig` presets that select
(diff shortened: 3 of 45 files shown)


rxbynerd:
Follow-up issues filed for the deferred review findings and verification observations: #468 (timeout error sentinel), #469 (misleading ctx-cancellation diagnostic — found during E2E verification, pre-existing), #470 (log on MaxTimeout clamp), #471 (span names from phase constants), #472 (ExitCode zero-value ambiguity doc), #473 (container partial output on timeout), #474 (OTel per-hook detail + span assertions), #475 (capability-derived exec check), #476 (job-path done-event integration test).

Issue #245, opened 2026-05-21
CLI: --dry-run with preflight probes (separate from --output-runconfig)


rxbynerd:
## Background

Issue #240 introduces `--output-runconfig <path>` on `stirrup harness`, which writes the resolved `RunConfig` to a file and exits without running the agentic loop. That flag is a *config capture* mechanism — it answers "what would have run?".

A separate but related need is a true *pre-flight check*: did the operator wire credentials correctly? Are the MCP servers reachable? Does the Cedar policy parse? `ValidateRunConfig` catches structural shape issues, but not these side-channel concerns — those only surface mid-run, often after a slow first turn has already burned cost.

## Request

Add a `--dry-run` flag to `stirrup harness` that runs the resolved `RunConfig` through every initialisation step *short of* the first agentic turn, then exits. The flag is independent of `--output-runconfig` (the two can be combined: `--dry-run --output-runconfig=resolved.json`).

Specifically, a dry-run exercises:

1. **`RunConfig` validation** — `ValidateRunConfig` (already exists, fast).
2. **Component construction** — `core.BuildLoop` builds every component (provider adapter, executor, edit strategy, verifier, permission policy, transport, trace emitter, guard rail). Construction failures surface here.
3. **Credential resolution** — `SecretStore.Resolve` for every `secret://` reference. STS / OIDC token exchanges for WIF credential types. Catches missing env vars and malformed federation rules.
4. **Provider reachability** — a short, low-cost ping to the provider (Anthropic: `GET /v1/models`, OpenAI: `GET /v1/models`, Bedrock: `ListFoundationModels`, Gemini: a minimal `streamGenerateContent` head request, etc.). Pings respect `--provider-retry-*` policy.
5. **MCP server reachability** — `initialize` handshake against every configured MCP server. Failures listed but not fatal unless `--strict` is also set.
6. **Cedar policy parse** — when `permissionPolicy.type=policy-engine`, parse and lint the policy file via the existing `policyengine` package.
7. **Container engine reachability** — when `executor.type=container`, ping the Docker / Podman socket and verify the requested image is pullable (without pulling).
8. **Workspace export destination probe** — when `executor.workspaceExportTo` is a `gs://` URI, verify the credential can list the bucket (HEAD on bucket metadata).
9. **Trace emitter probe** — for `traceEmitter.type=otel`, attempt one OTLP export with an empty batch; for `type=gcs`, verify bucket access; for `type=jsonl`, verify the directory is writable.

[...]

PR #488, opened 2026-07-11
ci: re-enable eval gate (cheap-model push gate + release model sweep) and add -race

rxbynerd:
## What

Fixes the two CI defects from the v0.1 test/CI review:

1. **The eval gate has been silently disabled since b472687** — its `if:` contained `… && github.event_name == 'disabled'`, so every run since 2026-06-30 reported SKIPPED while the docs claimed it ran on every push (live proof: PR #451's checks).
2. **CI never ran `go test -race`**, leaving the goroutine-heavy executor/egressproxy/provider/core packages unchecked for data races.

## Gate design

**Push gate (`ci.yml::eval-gate`)** — re-enabled on every push, every branch:
- Pinned to **Claude Haiku 4.5** (`claude-haiku-4-5-20251001`) via a new `stirrup-eval run --model` flag, which forwards `--model` to every harness invocation. The harness's `Changed()`-gated flag resolution means it overrides both the harness default and any suite-pinned `model_router` (verified against `harness.go`'s `applyOverrides`, plus new runner tests and a live fake-harness argv capture).
- **Only baselined suites execute.** `eval/suites/` now holds 4 suites but only `dogfood-seed` has a baseline; running `guardrail` (needs local vLLM), `tooluse`, and the openai-responses pin would spend tokens and produce failure noise while `compare` enforces nothing for them. Committing a baseline auto-enrols a suite.
- **Secret guard:** the existing check-creds step gains a repository check on its WIF leg. The Anthropic federation rule pins `assertion.repository`, so a fork's OIDC token passes the old `ACTIONS_ID_TOKEN_REQUEST_URL` probe but fails at exchange time — turning "skip gracefully" into per-task auth errors and a spurious red gate. Forks (and dependabot-actor pushes without secrets) now skip with a warning; a fork with its own `ANTHROPIC_API_KEY` still runs.

**Release sweep (`release.yml::eval-extended`)** — new matrix job running the baselined suites against **Claude Sonnet 5** and **Claude Opus 4.8** on every release tag. Wired exactly like the warn-mode vulnerability-gate: **not in `release.needs`**, so the non-blocking posture is structural — a regression paints the matrix cell red on the release run without blocking the release. Promotion to blocking is a one-line `needs:` change. `fail-fast: false` keeps per-model verdicts independent.

**OpenAI (GPT 5.6) left out deliberately:** neither ci.yml nor release.yml references an OpenAI credential today, so there is no auth path for it; adding one is a separate decision (the smoke-openai-wif workflow's repo-var gating could be reused if wanted).

**Caveat:** if the Anthropic-side federation rule constrains the OIDC `ref` claim to `refs/heads/main`, branch pushes and tag refs will fail the exchange. The smoke-anthropic workflow is `workflow_dispatch` (any-ref) with the same rule, so the rule is likely repo-pinned only — but if eval jobs go red with auth errors on non-main refs, widen the rule in the Anthropic console rather than weakening the jobs.

## Baselines: NOT refreshed — manual run needed

No Anthropic credentials are available in this environment, so `eval/baselines/dogfood-seed.json` is untouched: it still records the 2026-05-28 `claude-sonnet-4-6` run (all 5 tasks pass). The baseline's *shape* (per-task expected outcomes) is model-agnostic and the tasks are easy, so Haiku 4.5 is expected to pass — but the first gated push is the real test. To refresh under the gate's actual model:

```sh
just build
ANTHROPIC_API_KEY=... ./stirrup-eval run \
  --suite eval/suites/dogfood-seed.hcl \
  --harness "$PWD/stirrup" \
  --model claude-haiku-4-5-20251001 \
  --output eval/results/dogfood-seed
# then minimise result.json (outcomes only, durationMs: 0, update the
# _comment_ provenance line) and commit it as
# eval/baselines/dogfood-seed.json
```

## -race in CI

`_verify.yml` gains a scoped step — `go test -race ./harness/internal/executor/... ./harness/internal/provider/... ./harness/internal/core/...` (~9 s wall locally) — with a matching `just test-race` recipe. `go.work.sum` is refreshed because the step runs in workspace mode from the repo root and the recent dependabot bumps left the workspace sum file behind.

## Docs truth

`eval/suites/README.md`, `dogfood-seed.hcl`, `docs/eval.md`, `docs/deployment.md`, and `CONTRIBUTING.md` now describe the actual behaviour (cheap gate on push over baselined suites, extended sweep on release, WIF auth). Also corrected a claim that was never true: `publish-container` needs only `verify` — the eval gate never blocked the publish.

## Coordination with PR #480

PR #480 (build/sandbox-image) also touches ci.yml (adds `publish-sandbox-container`) and release.yml (sandbox publish + `release.needs`). This branch is off current main and confines its edits to the eval-gate job, the Verify test steps, and the new `eval-extended` job — whichever PR merges second needs a trivial rebase.

## Verification

- `just build`, `just test`, `just lint` — pass, 0 issues.
- `just test-race` — pass (~9 s).
- `actionlint` on all three workflows — clean apart from the known pre-existing SC2094 infos in the release aggregation script.
- New runner tests pin `--model` forwarding and the unset-flag legacy shape; `stirrup-eval run --model … --dry-run` and a live fake-harness run confirmed `--model claude-haiku-4-5-20251001` reaches the harness argv end-to-end.

### DIFF
--- a/.github/workflows/_verify.yml
+++ b/.github/workflows/_verify.yml
@@ -73,6 +73,15 @@ jobs:
         working-directory: eval
         run: go test ./...
 
+      # Race-detector pass over the goroutine-heavy packages (executor
+      # — which includes egressproxy — provider, and core). Scoped
+      # rather than ./... because the full sweep under -race is much
+      # slower for little extra concurrency coverage; the targeted set
+      # adds only a few seconds. Keep the package list in sync with
+      # the `test-race` recipe in the justfile.
+      - name: Run race detector on concurrency-heavy packages
+        run: go test -race ./harness/internal/executor/... ./harness/internal/provider/... ./harness/internal/core/...
+
       - name: Build stirrup binary
         working-directory: harness
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,17 @@ jobs:
     name: Eval Gate
     runs-on: ubuntu-latest
     needs: verify
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main' && github.event_name == 'disabled'
+    # Runs on every push (this workflow's only trigger), on every
+    # branch: the gate is the cheap-model regression check for the
+    # baselined suites, so PR branches get the signal before merge
+    # rather than discovering a regression on main. It was disabled
+    # outright between b472687 and this job's re-enable; the model is
+    # pinned to Haiku 4.5 (EVAL_MODEL below) to keep per-push cost
+    # low — the stronger-model sweep lives in
+    # release.yml::eval-extended. Missing credentials (fork clones,
+    # dependabot-actor pushes) skip gracefully via the check-creds
+    # step rather than failing.
+    #
     # id-token: write is required for Anthropic Workload Identity
     # Federation: actions/checkout's runner exposes
     # ACTIONS_ID_TOKEN_REQUEST_URL / ACTIONS_ID_TOKEN_REQUEST_TOKEN
@@ -87,6 +97,16 @@ jobs:
         shell: bash
         env:
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          # The WIF federation rule pins assertion.repository on the
+          # Anthropic side, so an OIDC token minted for a fork is
+          # rejected at exchange time. Without this guard a fork
+          # running its own Actions would see
+          # ACTIONS_ID_TOKEN_REQUEST_URL set, pass the check below,
+          # and then fail every task with a buried auth error — the
+          # exact failure mode this step exists to prevent. A fork
+          # that configures its own ANTHROPIC_API_KEY secret still
+          # runs the gate via the static-key leg.
+          WIF_ELIGIBLE: ${{ github.repository == 'rxbynerd/stirrup' }}
         run: |
           # Check whether a real credential source is available.
           # GitHub Actions sets ACTIONS_ID_TOKEN_REQUEST_URL only when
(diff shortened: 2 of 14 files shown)


rxbynerd:
**Eval Gate failure root-caused and fixed** (commit 4bc0219; green run: [29170513536](https://github.com/rxbynerd/stirrup/actions/runs/29170513536)).

**Root cause** — the flagged WIF caveat, now confirmed with hard evidence: the Anthropic federation rule refuses non-main-ref OIDC tokens. The failing run's per-task stderr showed every harness invocation dying in ~200 ms with `Anthropic WIF: token exchange returned 401 … authentication_error` (branch ref `ci/eval-gate-and-race`), while the only smoke-anthropic run (same rule, `main` ref, 2026-05-09) succeeded. The refused exchange then masqueraded as five bogus `pass → fail` "regressions" in the compare step. The refreshed Haiku baseline (6b8919e) was never the suspect.

**Fix** — a `stirrup harness --dry-run` preflight (the #245 preflight: real credential exchange, models-endpoint probe, zero completion tokens) now runs between the credential check and the suite run, with ref-sensitive semantics — deliberately *not* a blanket skip-on-failure:

| Ref | Refused exchange |
|---|---|
| non-main push | Skip live run + `::warning` naming the operator action (matches the fork/no-credential skip) |
| `refs/heads/main` | **Job fails** with an explicit "infrastructure failure, NOT an eval regression" error — main is the ref the rule is known to serve, so a refusal there must stay red |
| release tags (`eval-extended`) | **Matrix cell fails** with a legible error (job is structurally non-blocking; a green-but-vacuous sweep would misread as "evals passed") |

Run failures and baseline regressions still fail the gate on every ref. Verified live: the new run's preflight report names `[FAIL] provider-probe:anthropic: … token exchange returned 401` and the gate skips with the warning annotation instead of fabricating regressions.

**USER ACTION (out-of-repo, non-blocking for this PR):** to get live eval runs on branch pushes and release tags, widen federation rule `fdrl_01QJMFE86qFDdWzQXEekhYjY` in the Anthropic console to accept this repository's non-main refs (branches and `refs/tags/v*`). No workflow change is needed afterwards — branch/tag runs go live automatically. Until then: branches skip-with-warning, and the first push to main after merge exercises the full gate live (main-ref exchange is the known-good path).

